Issue #6, opened 2016-06-17
Include directories hard coded to authors own windows enviroment.


mdsitton:
I could fix the CMakeLists.txt and do a pull request but you'd need to change some stuff on your end. I have access to Linux, and OSX(though osx is a bit pointless if your only going to be targeting opengl 4.3? with its max of 4.1)

Also i'd like to contribute somewhat, but there isnt really any decent way to talk about the project maybe setup a https://gitter.im/ chat?

mdsitton:
I'm working on fixing this and will do a pull request.

VoyTechnology:
I agree. Gitter would be very useful. Asking questions with Github issues can be awkward

phase:
IRC is a possibility. esper or http://spi.gt/ are good.

DethRaid:
We have a discord at https://discordapp.com/invite/014ux6siiDogn9FLz

mdsitton:
Ok i've submitted my pull request #9.

DethRaid:
Pull request that fixed this is merged, so this issue should be solved.

PR #9, opened 2016-06-17
Clean up the cmake build system, and get it building on linux.

mdsitton:
Building on windows needs to be tested though. This will fix issue #6.


src/native/src/3rdparty/glad/include/KHR/khrplatform.h
```diff
@@ -26,7 +26,7 @@
```
mdsitton, line 26: Just fwi this commit also has a regenerated copy of glad with OpenGL 4.5 Core Profile + Extensions compatible with the core profile.

mdsitton:
There is also still some issues that need to be sorted out. For example if you want to attempt to run the genered binary you need to move it out of the src sub directory into top build folder (you should probably run cmake within the build folder but still is within a sub src folder)

Another issue currently it builds most of the files twice since it builds the shared library, and the test executable. Ideally we would want to have the test just link against the shared library, but i havent quite figured out how to do this yet.

I'd also like to move the files in git that are within the build directory somewhere else, then do a cmake custom command to copy them into the build folder after the build is done. But i will need to consult with @DethRaid where he would want to put them.

mdsitton:
Ok so most of that last comment is fixed. Only issue is i'd still like to move those files out of the build directory and maybe into sub project specific folders. so move build/test to src/test/runtime or something.

mdsitton:
This all still needs testing on windows though.

mdsitton:
For now i just moved the nova-test output directory to be inside the test folder so that nova-test can properly find the shaders.

mdsitton:
So currently the build of nova-test will not be relocatable outside of the build folder(on linux atleast). I'll need to setup the value INSTALL_RPATH so that the library can be found relative to the executable when "installing" though i haven't quite figured this out yet. Though this wont be relativent for the shared library though only the executable that links to it.

src/native/src/3rdparty/glad/include/KHR/khrplatform.h
DethRaid, line 26: Why did you regernate the loader?

mdsitton, line 26: I probably didnt need to, but the main reason was I saw you had some path changes in its headers. And just rather than figure out where things should have gone, i just regenerated it instead. Did you have any specific changes ontop of it that you needed?

DethRaid, line 26: Nothing specific. I was just wondering why you re-generated it.

mdsitton, line 26: It can also be a good idea to regenerate those types of things often, as they generally update the opengl xml spec quite a bit with bug fixes and new extensions and stuff.

src/native/src/cmake/FindGLFW3.cmake
```diff
@@ -0,0 +1,46 @@
...
+
+# default search dirs
+set( _glfw3_HEADER_SEARCH_DIRS 
+  "/usr/include"
+  "/usr/local/include"
+  "C:/Program Files (x86)/glfw/include" )
+set( _glfw3_LIB_SEARCH_DIRS
+  "/usr/lib"
+  "/usr/local/lib"
+  "C:/Program Files (x86)/glfw/lib-msvc110" )
```
DethRaid, line 22: I use CLion on top of win-builds. It works the best out of everything I've tried. By adding glfw as a subdirectory, like it was before, my compiler would build glfw, then Nova. No matter what compiler you generated the build files for, the compiler would be able to build glfw as needed. You've tied the Windows version to MSVC, which is not something we should do. A more optimal solution would be to not tie the Windows version to MSVC, and especially not to one specific version to MSVC.

DethRaid, line 22: I propose http://pastebin.com/k9R9ayJ3 as an alternate FindGLFW3.cmake. It checks first if your have the GLFW3 source code and, if so, it adds GLFW3 as a subdirectory. If you don't have the GLFW3 source code, it does things your way. I still don't like that, on Windows, you only search for binaries for MSVC11, but this works a bit better.

mdsitton, line 22: You don't need to compile glfw use the precompiled binaries from the website they contain various versions for different compilers http://www.glfw.org/download.html

You just need to rename the proper lib version for your compiler to lib instead of lib-compiler.

src/native/src/core/nova_renderer.cpp
```diff
@@ -6,8 +6,8 @@
 
 #include <easylogging++.h>
 
-#include "../gl/windowing/glfw_gl_window.h"
-#include "../utils/utils.h"
+#include "gl/windowing/glfw_gl_window.h"
```
DethRaid: Why did you do this?

src/native/src/cmake/FindGLFW3.cmake
mdsitton, line 22: And if you want to compile glfw i can configure cmake to do that if it doesnt find the pre-compiled version would that work for you?

src/native/src/core/nova_renderer.h
```diff
@@ -7,12 +7,13 @@
 
 #include <memory>
 
-#include "../interfaces/iwindow.h"
+#include "interfaces/iwindow.h"
+#include "mc/mc_objects.h"
+
 #include "nova.h"
-#include "../mc/mc_objects.h"
+#include "types.h"
```
DethRaid, line 14: Nope. types.h is for typedefs that make the code more readable. mc_objects.h is for C-style structs that the Nova DLL will export. Don't put mc_objects.h in types.h.

src/native/src/core/texture_manager.cpp
```diff
@@ -53,7 +53,7 @@ void texture_manager::add_texture(mc_atlas_texture & new_texture, atlas_type typ
     GLenum format = GL_RGB;
     switch(new_texture.num_components) {
         case 1:
-            format = GL_R;
+            format = GL_RED;
```
DethRaid, line 59: I did not know that they had changed the single-color symbols.

src/native/src/gl/objects/gl_shader_program.h
```diff
@@ -12,8 +12,8 @@
 #include <unordered_map>
 #include <vector>
 
-#include "../../interfaces/ishader.h"
-#include "../glad/glad.h"
+#include "interfaces/ishader.h"
```
DethRaid: Again with no leading "../"

src/native/src/cmake/FindGLFW3.cmake
DethRaid, line 22: I very much do not want to download a pre-build binary if at all possible. I've had too much trouble with compiler versions. It's much easier for me to clone a git repo and add it as a subdirectory.

I think my updated FindGLFW3.cmake will handle source code GLFW. You seem to know CMake better than me, though, so if you think there's a better way to do it go ahead.

src/native/src/core/nova_renderer.h
mdsitton, line 14: I didn't it looks like at some point i moved the imports around. mc_objects.h is above a few lines, right below iwindow.h

src/native/src/core/texture_manager.cpp
mdsitton, line 59: from here:
https://www.opengl.org/sdk/docs/man2/xhtml/glTexGen.xml

Specifies a texture coordinate. Must be one of GL_S, GL_T, GL_R, or GL_Q.

So GL_R had something to do with texture coords in old opengl versions.

As far as i can tell they never changed it. Its always been that way. The values of each are also different
GL_RED: 0x1903
GL_R: 0x2002

src/native/src/core/nova_renderer.cpp
mdsitton: I changed most of the relative includes to absolute from the src directory, unless they were files within the current folder. I cant stand seeing multiple levels of ../../../ in #include statments...

src/native/src/gl/objects/gl_shader_program.h
mdsitton: See other comment.

src/native/src/core/nova_renderer.cpp
DethRaid: Except that now it doesn't compile on my machine

src/native/src/core/nova_renderer.h
DethRaid, line 14: Huh. So it is.

Why'd you move the imports around?

mdsitton, line 14: I guess i probably just thought the includes from the current directory should be together?

DethRaid, line 14: Okay.

It doesn't really matter, I was just wondering why

mdsitton:
Rebased ontop of current master (thats why comments are a bit weird and out of place now)

I mentioned some of the issues with doing the GLFW build from our cmake.I've stayed up quite late and need to get to sleep. I'll implement the least annoying method when i wake up.

DethRaid:
Glancing at the code this looks good. I'll have to pull it locally and test it on my machine when I get home from work. How close to done do you think you are?

mdsitton:
Well currently i have an issue on Arch Linux that i'm not certain if its a problem with the cmake config or a bug with the build configuration on an arch library. So i need to spin up a VM of another distro and test it there to figure out which it is. It still needs to be tested on MinGW. I need to add the libarchive dependency still.
I wish i could also figure out the exporting issues on windows as well, but thats going to require a redesign of all of the classes. or to define a set of api classes that should be simple and doesn't have any STL stuff used externally.

So the main issues i can probably finish in a day. I'm off work for the weekend after tonight so i'll have more time to work on it again.

src/native/src/3rdparty/json/ChangeLog.md
```diff
@@ -0,0 +1,224 @@
+# Change Log
```
VoyTechnology, line 1: Instead of committing all the 3rdparty files, can git submodules not be used instead? This would remove the clutter that is created by constantly changing dependencies.

mdsitton, line 1: I was talking with dethraid about this. Personally I don't want to have the whole history of another repo included since it's not really needed, and it just increases the time that it takes onto clone the repo to be ready to build. That's the main reason really. But I haven't been including all dependencies this way. Just the single file style headers that you just throw in your project. Though I happened to fork one of them integrated some changes that never made it to the official repo.

VoyTechnology, line 1: I think longer clone times are better than littered source. As an alternative it can in the make file or some preparation script to clone the dependency with --depth=1 and .gitignore that directory so it's not accidentally added.

DethRaid, line 1: [Subtrees](http://blogs.atlassian.com/2013/05/alternatives-to-git-submodule-git-subtree/) look pretty good too. What do you guys think?

mdsitton, line 1: Just going to mention one more time that most of the dependencies will be downloaded by cmake and compiled at the first run if needed. The ones that are in-tree now are meant to live in your source tree and be compiled with your project.

PR #21, opened 2017-01-29
Implemented Mouse Handling, though no scrolling yet

Maximilian-Reuter:
this pr implements mouse handling, and as an added bonus, disabled buttons are actually drawn disabled


src/main/cpp/input/INPUTHANDLER.H
```diff
@@ -0,0 +1,36 @@
+#ifndef INPUT_HANDLER_H
+#define INPUT_HANDLER_H
+
+
+
+#include "GLFW/glfw3.h"
+#include <queue>
+#include <mutex>
+#include "../mc_interface/mc_objects.h"
+using namespace std;
```
DethRaid: I'd rather we not do this

src/main/java/com/continuum/nova/Mouse.java
```diff
@@ -0,0 +1,248 @@
...
+    public static void create() {
+        if (!created) {
+            initialize();
+            created = true;
+        }
+
+    }
+
+    public static boolean isCreated() {
+
```
DethRaid, line 89: So much whitespace...

src/main/java/com/continuum/nova/Mouse.java
```diff
@@ -0,0 +1,248 @@
+package com.continuum.nova;
+//
+// Source code recreated from a .class file by IntelliJ IDEA
```
DethRaid, line 3: Did you decompile the LWJGL mouse class? Why not just extend it and override its methods?

PR #27, opened 2017-02-03
Crash fix

tobiasschulz:


nova.sh
```diff
@@ -103,6 +103,51 @@ create_patch() {
     file patches/nova.patch | grep UTF-16 && (iconv --from-code UTF-16 --to-code UTF-8 patches/nova.patch > x && mv x patches/nova.patch)
 }
 
+tracing_nova_add() {
```
DethRaid, line 106: Tell me about this tracing thing. What is it and what does it do?

src/main/cpp/render/objects/gl_mesh.cpp
```diff
@@ -102,8 +102,8 @@ namespace nova {
                 glEnableVertexAttribArray(0);   // Position
                 glEnableVertexAttribArray(1);   // Texture UV
 
-                glVertexAttribPointer(0, 3, GL_FLOAT, GL_FALSE, 5 * sizeof(GL_FLOAT), (void *) 0);
-                glVertexAttribPointer(1, 2, GL_FLOAT, GL_FALSE, 5 * sizeof(GL_FLOAT), (void *) (3 * sizeof(GLfloat)));
+                glVertexAttribPointer(0, 3, GL_FLOAT, GL_FALSE, 5 * sizeof(GLfloat), (void *) 0);
```
DethRaid, line 105: Oops. Can't believe I did that. Good thing you cought it

src/main/cpp/render/objects/render_object.cpp
```diff
@@ -0,0 +1,24 @@
...
+ * \brief
+ *
+ * \author ddubois 
+ * \date 02-Feb-17.
+ */
+
+#include "render_object.h"
+
+namespace nova {
+    /*render_object::render_object(render_object&& other) {
```
DethRaid: You should delete this file (or I should really) before merging

DethRaid (changes requested):
What ended up being the problem?

src/main/cpp/render/objects/render_object.cpp
tobiasschulz: what is it for?

DethRaid: I thought I might have to write my own move constructor but I didn't have to

nova.sh
tobiasschulz: it inserts a debug message at the beginning of every method. I was unable to get a useful stack trace form the debugger, so that was the only thing that helped me here. It's hacky but it always works, also on linux without any debugger, and its only temporary

tobiasschulz:
- the real problem was the unsigned short thing
- then your fix was buggy because of that remove_if misunderstanding
- not using raw pointers for "render_object" and "gl_mesh" is cleaner
    - but using a smart pointer for "gl_mesh" ensures that nothing brakes if "render_object" is copied by value for whatever reason
- another problem is the inconsistency of the texture names. Maximilian suggested that "texture/" and ".png" should be removed before atlas lookup

DethRaid (approved):

PR #28, opened 2017-02-04
Window resize

Maximilian-Reuter:


jars/shaderpacks/default/shaders/gui.vert
```diff
@@ -4,11 +4,11 @@ layout(location = 0) in vec3 position_in;
 layout(location = 1) in vec2 uv_in;
 layout(location = 2) in vec3 color_in;
 
-layout(shared, std140) uniform gui_uniforms {
-    float viewWidth;
-    float viewHeight;
-    mat4 gbufferModelView;
+layout(binding = 0,std140) uniform gui_uniforms {
```
DethRaid, line 7: We shouldn't need to specify the binding here, Nova should pick it up at runtime. Plus, if you do specify the binding, you'll have to put that on the UBO every time it's used in a shader

src/main/cpp/render/objects/shaders/uniform_buffers_definitions.cpp
```diff
@@ -10,15 +10,15 @@
 
 namespace nova {
     el::base::Writer &operator<<(el::base::Writer &out, gui_uniforms uniforms) {
-        out << "gbufferModelView: " << uniforms.gbufferModelView
-            << " gbufferProjection:" << uniforms.gbufferProjection
+        out << "gbufferModelView: \n" << uniforms.gbufferModelView
```
DethRaid, line 13: I probably should have done this. All well.

src/main/cpp/render/uniform_buffer_store.cpp
```diff
@@ -69,7 +75,7 @@ namespace nova {
 
     void uniform_buffer_store::update_per_frame_uniforms() {
         // TODO: Fill in the per frame uniforms
-        per_frame_uniforms_buffer.send_data(per_frame_uniform_variables);
+        //per_frame_uniforms_buffer.send_data(per_frame_uniform_variables);
```
DethRaid, line 78: Why is this commented out

DethRaid (changes requested):
Fix the logging and I'll merge

src/main/cpp/render/uniform_buffer_store.cpp
Maximilian-Reuter: the per frame uniforms don't get any data set and they do produce an Invalid operation or Value down the line somewhere. until we actually need it, it just complicates things as opengl stuff has a tendency to explode in your face

DethRaid (approved):

PR #30, opened 2017-02-07
Working background

Maximilian-Reuter:


src/main/cpp/input/InputHandler.h
```diff
@@ -165,6 +167,8 @@ namespace nova {
 		struct mouse_button_event dequeue_mouse_button_event();
 		void queue_mouse_position_event(struct mouse_position_event  e);
 		struct mouse_position_event dequeue_mouse_position_event();
+        void queue_mouse_scroll_event(struct mouse_scroll_event  e);
```
DethRaid: You know you don't need to put `struct` here, right?

src/main/cpp/render/objects/gl_mesh.cpp
```diff
@@ -15,8 +15,8 @@ namespace nova {
 
     gl_mesh::gl_mesh(mesh_definition &definition) {
         create();
-        set_data(definition.vertex_data, definition.vertex_format, usage::static_draw);
-        set_index_array(definition.indices, usage::static_draw);
+        set_data(definition.vertex_data, definition.vertex_format, usage::dynamic_draw);
```
DethRaid: Why?

src/main/java/com/continuum/nova/NovaNative.java
```diff
@@ -11,6 +11,7 @@
     String GUI_ATLAS_NAME = "gui";
```
DethRaid: Should probably make these static... that's probably my fault

src/main/java/com/continuum/nova/NovaRenderer.java
```diff
@@ -68,7 +70,7 @@ public NovaRenderer() {
     public void onResourceManagerReload(@Nonnull IResourceManager resourceManager) {
         this.resourceManager = resourceManager;
 
-        if(firstLoad) {
+        if (firstLoad) {
```
DethRaid, line 73: Reformatting boo

src/main/java/com/continuum/nova/NovaRenderer.java
```diff
@@ -273,26 +282,32 @@ private void updateWindowSize(){
...
     }
 
     public void updateCameraAndRender(float renderPartialTicks, long systemNanoTime, Minecraft mc) {
-        if(NovaNative.INSTANCE.should_close()) {
+        if (NovaNative.INSTANCE.should_close()) {
             Minecraft.getMinecraft().shutdown();
         }
+
+        if (Minecraft.getMinecraft().currentScreen != null) {
+
```
DethRaid, line 303: Why is there an empty if block here?

src/main/java/com/continuum/nova/NovaRenderer.java
```diff
@@ -273,26 +282,32 @@ private void updateWindowSize(){
...
+        }
+        NovaDraw.novaDrawScreen(Minecraft.getMinecraft().currentScreen, renderPartialTicks);
+
         NovaNative.INSTANCE.execute_frame();
         updateWindowSize();
-        int scalefactor = new ScaledResolution(Minecraft.getMinecraft()).getScaleFactor()*2;
-        if (scalefactor!=this.scalefactor) {
+        int scalefactor = new ScaledResolution(Minecraft.getMinecraft()).getScaleFactor() * 2;
```
DethRaid, line 310: The one time I like your reformatter

src/main/java/com/continuum/nova/gui/NovaDraw.java
```diff
@@ -62,11 +76,12 @@ public static void draw(ResourceLocation texture, Integer[] indexBuffer, Float[]
      * @param vertices    the vertices as Vertex objects
      */
     public static void draw(ResourceLocation texture, Integer[] indexBuffer, Vertex[] vertices) {
+
         Float[] vertexbuffer = new Float[vertices.length * 9];
         for (int v = 0; v < vertices.length; v++) {
             vertexbuffer[v * 9] = vertices[v].x;
             vertexbuffer[v * 9 + 1] = vertices[v].y;
-            vertexbuffer[v * 9 + 2] = vertices[v].z;
+            vertexbuffer[v * 9 + 2] = currentZ;
```
DethRaid, line 84: Since we're not getting the z from `Vertex` objects, let's cascade this change throughout NovaDraw

src/main/java/com/continuum/nova/gui/NovaDraw.java
```diff
@@ -100,33 +111,41 @@ public static void draw(ResourceLocation texture, Integer[] indexBuffer, Vertex[
      * @param texWidth  texture / UV coordinates, relative to the original minecraft textures (not the texture atlas)
      * @param texHeight texture / UV coordinates, relative to the original minecraft textures (not the texture atlas)
      */
-    public static void drawRectangle(ResourceLocation texture, Rectangle2D.Float rect, float z, Rectangle2D.Float textureCoords) {
+    public static void drawRectangle(ResourceLocation texture, Rectangle2D.Float rect, float z, Rectangle2D.Float textureCoords,Color vertexColor) {
```
DethRaid: No need to pass in Z here, since we're not using the passed in Z value anywhere

src/main/java/com/continuum/nova/gui/NovaDraw.java
```diff
@@ -155,19 +174,25 @@ private static void computeCorrectMousePosition() {
...
+            NovaNative.INSTANCE.send_gui_buffer_command(command);
+            long end = System.nanoTime();
+            LOG.info("time used to copy buffers to c++ : " + (end - timePrev) + "time used to alloc buffers and fill: "+((end - timeWithAlloc) - (end - timePrev)));
+            Memory.purge();
         }
+
+
     }
 
     public static class Vertex {
```
DethRaid, line 194: Please remove the z attribute from this class

DethRaid (changes requested):
Please delete commented out code and cascade your changes so that `Vertex` doesn't have a z value

src/main/cpp/input/InputHandler.h
Maximilian-Reuter: hmm okay was told that was neccessary

DethRaid: Nope! http://stackoverflow.com/questions/15181765/passing-structs-to-functions

In C, maybe

src/main/java/com/continuum/nova/NovaRenderer.java
tobiasschulz: why do you dislike these spaces so much? I can't think of anyone or any project that uses this particular code style...

DethRaid: My code style is weird. I used to use the Vertical For Loop

```
for( int i = 0;
    i < x;
    i++ );
```

Then I moved on to super-spaced out everything

```
void foo( int x );
```

But I've since abandoned both of those.

I like not having spaces before parens in method calls and declarations and for/if/while statements is that this way it feels like the things inside the parens are more connected to the function/if/for/whatever. I recognize that this isn't what many people do, though, so I won't enforce this on your code. I booed on this line because it was my code, but had been reformatted. I'd rather we not reformat each other's code (and if I reformat yours, please call me out on it)

DethRaid: (I don't see a way to edit that comment. Hopefully you get the idea)

src/main/java/com/continuum/nova/NovaNative.java
```diff
@@ -11,7 +11,6 @@
...
     String BLOCK_COLOR_ATLAS_NAME = "block_color";
     String FONT_ATLAS_NAME = "font";
-    String OPTIONS_BACKGROUND = "options_background";
```
DethRaid: Why was this line removed?

Maximilian-Reuter: because it is now in free textures and does not need its own atlas anymore

src/main/java/com/continuum/nova/NovaRenderer.java
Maximilian-Reuter: it's not empty anymore

src/main/java/com/continuum/nova/NovaNative.java
Maximilian-Reuter: that would be redundant as it is an Interface

src/main/cpp/render/objects/gl_mesh.cpp
Maximilian-Reuter: i wanted to find a way to reuse meshes and their buffers but didn't get anywhere. Reverted.

DethRaid (approved):

PR #31, opened 2017-02-08
added modelmatrix to perform transformation on vertices

Maximilian-Reuter:


src/main/java/com/continuum/nova/gui/NovaDraw.java
```diff
@@ -30,6 +32,9 @@
 
     private static float currentZ;
 
+    private static Mat4 ModelMatrix = new Mat4();
```
DethRaid, line 35: Static variables are usually UPPER_SNAKE_CASE in Java...

Maximilian-Reuter: Nope that only applies to constants

Maximilian-Reuter: if anything it should be modelMatrix

PR #45, opened 2017-06-22
Fix The Shaders And Includes

ckissane:
Did It!


src/main/java/Start.java
```diff
@@ -8,8 +8,7 @@
 {
     public static void main(String[] args)
     {
-        System.out.println("Running Nova");
-        Main.main(concat(new String[] {"--version", "mcp", "--accessToken", "0", "--assetsDir", "assets", "--assetIndex", "1.10", "--userProperties", "{}"}, args));
+        Main.main(concat(new String[] {"--version", "mcp", "--accessToken", "0", "--assetsDir", "assets", "--assetIndex", "1.8", "--userProperties", "{}"}, args));
```
DethRaid: Please undo this change. Asset index 1.10 is the correct one

DethRaid (commented):
Did you run the nova-test executable to ensure that you didn't break anything?

DethRaid (commented):
Why did you change the names of the shaders?

ckissane:
nothing is broken

src/main/java/Start.java
ckissane: hmm I dont know how that happend

PR #47, opened 2017-06-22
Water Shaders

ckissane:
Added Water Shaders


src/main/cpp/render/nova_renderer.cpp
```diff
@@ -66,6 +66,8 @@ namespace nova {
 
         render_gbuffers();
 
+        //render_gbuffers();
```
skmedix, line 69: But why? It's just a useless comment

src/main/java/Start.java
```diff
@@ -8,7 +8,6 @@
 {
     public static void main(String[] args)
     {
-        System.out.println("Running Nova");
```
skmedix, line 11: Same here

PR #48, opened 2017-06-23
Removal Of inner faces of transparent groups

ckissane:


jars/config/shaders.json
```diff
@@ -74,6 +74,11 @@
             "fallback": "gbuffers_terrain"
         },
         {
+            "name": "gbuffers_lava",
```
DethRaid: Why did you add a lava shader? There shouldn't be a lava shader

jars/shaderpacks/default/shaders/gbuffers_lava.fsh
```diff
@@ -0,0 +1,78 @@
+#version 450
```
DethRaid: Please remove this shader

jars/shaderpacks/default/shaders/gbuffers_lava.vsh
```diff
@@ -0,0 +1,68 @@
+#version 450
```
DethRaid: Please remove this shader

src/main/cpp/render/nova_renderer.cpp
```diff
@@ -94,6 +94,8 @@ namespace nova {
         render_shader(terrain_shader);
         auto& water_shader = loaded_shaderpack->get_shader("gbuffers_water");
         render_shader(water_shader);
+        auto& lava_shader = loaded_shaderpack->get_shader("gbuffers_lava");
```
DethRaid: Again, please remove the lava shader

DethRaid (changes requested):
Please remove the lava shader. It is not part of Optifine Shaders and thus should not be a part of Nova

DethRaid (approved):

PR #54, opened 2017-08-03
Added Linux patch options

sayomaki:
Support 'applyPatch' for linux using 'patch' command


tobiasschulz:
For windows there is ${project.projectDir} in both paths. For linux the relative paths are just hardcoded. Are you sure that works?

tobiasschulz (changes requested):
add ${project.projectDir} for linux

tobiasschulz (changes requested):
Add ${project.projectDir} for linux

sayomaki:
Well technically it was using relative paths in nova.sh also, which is the current install script

tobiasschulz (approved):

PR #59, opened 2017-08-09
Fix MCP download error

Flandyn:
Fixes the "Exception calling "DownloadFile" with "2" argument(s): "The remote server returned an error: (404) Not Found."" error that occurs trying to download the MCP

### DIFF
--- a/nova.ps1
+++ b/nova.ps1
@@ -82,7 +82,7 @@ function New-NovaEnvironment {
 
     # Download MCP
     $wc = New-Object System.Net.WebClient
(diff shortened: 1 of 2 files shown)


DethRaid (commented):
Do you know if the mcp website will keep this new url? I don't want to have to change this again in a couple weeks

foxfirecodes:
If you do go through with this PR, don't forget to change the Unix installer script too.

Flandyn:
According to the website this is temporary, however i have no idea how long it will last and it's annoying having to change the link every time i clone the repository

Issue #86, opened 2017-10-20
Cannot compile Nova


sambow23:
Every time I run ".\gradlew build buildDll", I get this

`Configuring for MSVC 64-bit
.\gradlew : C:\Users\Sam Bowman\nova-renderer\src\main\java\net\minecraft\client\renderer\VertexBuffer.java:3: error: package 
com.google.common.primitives does not exist
At line:1 char:1
+ .\gradlew build buildDll
+ ~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (C:\Users\Sam Bo... does not exist:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
 
import com.google.common.primitives.Floats;
                                   ^
C:\Users\Sam Bowman\nova-renderer\src\main\java\net\minecraft\client\renderer\VertexBuffer.java:15: error: package 
org.apache.logging.log4j does not exist
import org.apache.logging.log4j.LogManager;
                               ^
C:\Users\Sam Bowman\nova-renderer\src\main\java\net\minecraft\client\renderer\VertexBuffer.java:16: error: package 
org.apache.logging.log4j does not exist
import org.apache.logging.log4j.Logger;
                               ^
C:\Users\Sam Bowman\nova-renderer\src\main\java\net\minecraft\util\math\BlockPos.java:3: error: package com.google.common.collect does not 
exist
import com.google.common.collect.AbstractIterator;
                                ^
C:\Users\Sam Bowman\nova-renderer\src\main\java\net\minecraft\util\math\BlockPos.java:4: error: package com.google.common.collect does not 
exist
import com.google.common.collect.Lists;
                                ^
C:\Users\Sam Bowman\nova-renderer\src\main\java\net\minecraft\util\math\BlockPos.java:7: error: package javax.annotation.concurrent does 
not exist
import javax.annotation.concurrent.Immutable;
                                  ^
C:\Users\Sam Bowman\nova-renderer\src\main\java\net\minecraft\util\math\BlockPos.java:10: error: package org.apache.logging.log4j does not 
exist
import org.apache.logging.log4j.LogManager;
                               ^
C:\Users\Sam Bowman\nova-renderer\src\main\java\net\minecraft\util\math\BlockPos.java:11: error: package org.apache.logging.log4j does not 
exist
import org.apache.logging.log4j.Logger;
                               ^
[...]

PR #87, opened 2017-10-21
Added missing dependency that required for build

Plasticable:
Added missing dependency that adds `javax` annotations which required for some minecraft source files.
You can see errors at #86 like this 
```
AxisAlignedBB.java:4: error: cannot find symbol
import javax.annotation.Nullable;
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -123,6 +123,8 @@ dependencies {
         )
     }
 
+    compile group: 'com.google.code.findbugs', name: 'jsr305', version: '3.0.0'
+	
     // Add all the dependencies that Minecraft has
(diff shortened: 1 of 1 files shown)


DethRaid:
No one else has had a problem with this. Why do we need to add it in now?

Plasticable:
At least, i had that problem too. Can you say, from what jar loaded `javax.annotation.*` classes in your project?

DethRaid:
There's a chance that I have that annotation from IntelliJ. After doing some research, it looks like not having this annotation is a reasonably common thing. Alright, I'll accept it

PR #103, opened 2018-01-24
Now we copy java files!

ckissane:


DethRaid (changes requested):
You're copying the files into src/main/Java/mcp... You realize that all the Java code looks for the Minecraft files in src/main/Java/net though, right? That Minecraft code has to end up in that directory eventually

Did you even test this?

ckissane:
I did test

ckissane:
Only errors now arise from patch not being applied (mc compiles fine, Nova hooks don't)

DethRaid:
Wasn't the whole point of this to make the patch file apply properly?

DethRaid (changes requested):
We need this code. You can't just comment it out, it's integral to having Minecraft's lighting work

ckissane:
I put the code back...
The patch works,
The code doesn't compile though.

ckissane:
K it builds and runs now

DethRaid (approved):

PR #130, opened 2018-07-14
Mouse and ub fix

DethRaid:


src/main/java/com/continuum/nova/NovaRenderer.java
```diff
@@ -278,6 +286,47 @@ public void preInit() {
         chunkUpdateListener = new ChunkUpdateListener(chunksToUpdate);
     }
 
+    private void installNative() throws IOException {
+        if((Boolean) Launch.blackboard.get("fml.deobfuscatedEnvironment")) {
+            LOG.info("Nova is very likely running in a development environment, trying to load native from run directory...");
+            try {
+                if (Platform.isWindows()) {
```
DethRaid, line 293: I keep asking this and you keep not answering: Why are we doing this ourselves instead of letting JNA handle loading the native DLL?

src/main/java/com/continuum/nova/NovaRenderer.java
```diff
@@ -278,6 +286,47 @@ public void preInit() {
...
+        if(!novaConfigDir.exists() || !novaConfigDir.isDirectory()) {
+            if(!novaConfigDir.mkdirs()) {
+                throw new IOException("Failed to create directory " + novaConfigDir.getAbsolutePath());
+            }
+        }
+
+        File toLoad;
+        if(Platform.isWindows()) {
+            if(Platform.is64Bit()) {
+                toLoad = Native.extractFromResourcePath("/nova-renderer-x64.dll");
```
DethRaid, line 315: Again, JNA handles this for us.

src/main/java/com/continuum/nova/transformer/NovaClassTransformer.java
```diff
@@ -81,7 +81,9 @@ public MethodVisitor visitMethod(int access, String name, String desc,
                 }
 
                 private boolean removeCalls(int opcode, String owner, String name, String desc, boolean itf) {
-                    if (owner.equals("org/lwjgl/opengl/Display") && (name.equals("create") || name.equals("getDisplayMode"))) {
+                    if (owner.equals("org/lwjgl/opengl/Display") &&
+                            (name.equals("create") || name.equals("getDisplayMode") || name.equals("getWidth") || name.equals("getHeight")
```
DethRaid, line 85: Are you sure this is fine? Will it not call into an uninitialized Display instance and cause problems?

Those functions are called by MC (or there wouldn't be any need to leave them alone). `getWidth` and `getHeight` can be redirected to get the width and height of the GLFW window, `getDisplayMode` we can probably do something useful for, but we should be fine with making `create` a no-op.

However... these methods are called by Minecraft as it tries to use LWJGL's Display class. Seems to me we'd get more mileage out of refactoring Minecraft to not use the LWJGL Display. Is there too much BS with mixins not being good at modifying methods to make that viable?

Niko-sk2x: It was late and I was trying to get mouse working, so some of them areunnecessary, but also won't hurt.

iirc, doing it for `create` broke this code because it throws LWJGLException, and the code fails because of that somehow. And nova already replaces places where `create` is called. There is just one place.

`getDisplayMode` - this one  just returns a field from Display, not sure if it's necessary.
`getWidth`, `getHeight` - I thought here they actually return some sensible value, and that they are used by MC. Looking at the code, they return 0 anyway (they just return a field from Display)
`isActive` - this is the one **actually needed fopr things to work**. This one calls into native code to check if the current window is active, and very surprisingly, it **actually works with nova anyway**. 

And if we ever want to replace any of these with mixin, we would need the method call to be there to target it with mixin.

The only issue ith mixin is that you can't just tell it "target all methods in all classes", you can with a transformer. Mixins are good for local changes. They are terrible for general "replace all of these with these" changes.

PR #136, opened 2018-07-17
Feature/cibuild

Janrupf:


mod/build.gradle
```diff
@@ -105,7 +150,319 @@ processResources {
     from(sourceSets.main.resources.srcDirs) {
         exclude "mcmod.info"
     }
+
+    exclude "META-INF/services/"
+    exclude "META-INF/maven/"
+}
+
+task packageDefaultShaderpack(type: Zip) {
+    from fileTree(dir: "run/shaderpacks/DefaultShaderpack")
```
DethRaid, line 159: It was called `default` before. Why the change?

mod/build.gradle
```diff
@@ -105,7 +150,319 @@ processResources {
...
+    props.load(tmpBuildPropsFile.newReader())
+    return props
+}
+
+void saveTmpBuildProperties(Properties props) {
+    ensureFile(tmpBuildPropsFile)
+    props.store(tmpBuildPropsFile.newWriter(false), "MACHINE GENERATED FILE, DO NOT EDIT")
+}
+
+void buildRelease(String platform) {
```
DethRaid: This should be three functions: one for letting CMake decide, one for Windows, one for Linux. We also should not use a regular string for the platform - there will be three functions, so we can just call those and not use a string as an enum

mod/build.gradle
```diff
@@ -105,7 +150,319 @@ processResources {
...
+    } else {
+        include "libnova-renderer.so"
+    }
 }
 
 runClient.outputs.upToDateWhen { false }
-runClient.standardOutput = System.out
\ No newline at end of file
+runClient.standardOutput = System.out
+//noinspection GroovyAssignabilityCheck
```
DethRaid, line 449: Delete, don't comment out

mod/run/config/forge.cfg
```diff
@@ -0,0 +1,84 @@
+# Configuration file
```
DethRaid: What is in this file that we need?

mod/run/config/forgeChunkLoading.cfg
```diff
@@ -0,0 +1,51 @@
+# Configuration file
```
DethRaid: What is in this file that we need?

mod/run/config/nova/config.json
```diff
@@ -1,6 +1,6 @@
 {
   "settings": {
-    "loadedShaderpack": "default",
+    "loadedShaderpack": "DefaultShaderpack",
```
DethRaid, line 3: Why the change?

mod/run/config/nova/logging.conf
```diff
@@ -10,11 +10,11 @@
     FORMAT                  = [%datetime{%h:%m:%s}] [Nova Renderer/%level]: %fbase %line - %msg
     FILENAME                = "logs/nova.log"
     ENABLED                 = true
-    TO_FILE                 = true
+    TO_FILE                 = false
```
DethRaid: Why?

When the log goes to file, it becomes persistent and easily searchable and awesome. When it does not, it gets lost when we run again - plus it's not easily searcahble, or easily manipulated...

Why did you make this change?

mod/run/config/nova/logging.conf
```diff
@@ -10,11 +10,11 @@
...
     TO_STANDARD_OUTPUT      = true
 * TRACE:
     FORMAT                  = [%datetime{%h:%m:%s}] [Nova Renderer/%level]: %fbase %line - %msg
     FILENAME                = "logs/nova.log"
     ENABLED                 = true
-    TO_FILE                 = true
-    TO_STANDARD_OUTPUT      = true
\ No newline at end of file
+    TO_FILE                 = false
```
DethRaid: Same as the other comment in this file

mod/run/config/splash.properties
```diff
@@ -0,0 +1,17 @@
...
+memoryGood=0x78CB34
+font=0x0
+barBackground=0xFFFFFF
+barBorder=0xC0C0C0
+memoryLow=0xE42F2F
+rotate=false
+memoryWarn=0xE6E84A
+showMemory=true
+bar=0xCB3D35
+enabled=true
```
DethRaid: Enabled needs to be false, or the splash screen will make a bunch of GL calls and crash immediately

mod/settings.gradle
```diff
@@ -0,0 +1 @@
+rootProject.name = "Nova_Renderer"
```
DethRaid: Why is there an underscore?

mod/src/main/java/com/continuum/nova/NovaRenderer.java
```diff
@@ -291,51 +290,54 @@ private void installNative() throws IOException {
...
+            if(NovaRenderer.class.getResource("/libnova-renderer.so") == null) {
+                // User friendly fail, show dialog box before crashing
+                JOptionPane.showMessageDialog(null, "The nova build you are using was build without Linux support!\n" +
+                        "   Please download an official build from our buildserver!");
+                throw new IllegalStateException("Linux is not supported by the current nova build");
             }
+            Utils.extractResource("/libnova-renderer.so", nativeExtractDir.toPath(), true);
+            Utils.extractResource("/libnova-profiler.so", nativeExtractDir.toPath(), true);
         }
+        nativeExtractDir.deleteOnExit();
```
DethRaid, line 323: How much time does this add to startup? Seems to me that extracting a multi-megabyte file will add a non-trivial cost to an already long startup

mod/src/main/java/com/continuum/nova/mixin/MixinMinecraft.java
```diff
@@ -294,7 +308,11 @@ private void init() throws LWJGLException {
         bar.step("Loading Model Manager");
         this.modelManager = new ModelManager(this.textureMapBlocks);
         this.blockColors = BlockColors.init();
-        NovaRenderer.getInstance().loadShaderpack("default", this.blockColors);
+        try {
+            NovaRenderer.getInstance().loadShaderpack(Utils.getShaderpackNameFromConfig(), this.blockColors);
+        } catch (IOException e) {
+            throw new LoaderExceptionModCrash("Nova failed to read shaderpack name from config", e);
```
DethRaid: We don't need to crash - we know the name of the default shaderpack, so why not load that?

mod/src/main/java/com/continuum/nova/mixin/MixinSplashProgress.java
```diff
@@ -0,0 +1,18 @@
...
+import org.spongepowered.asm.mixin.injection.Redirect;
+
+@Mixin(SplashProgress.class)
+public class MixinSplashProgress {
+    @Redirect(
+            method = "start",
+            at = @At(target = "getBool", value = "INVOKE"),
+            remap = false
+    )
+    private static boolean splashEnabled() {
```
DethRaid, line 15: Why are we doing this with a mixin and not with the config file?

mod/src/main/java/com/continuum/nova/NovaRenderer.java
```diff
@@ -40,12 +38,12 @@
 
 import javax.annotation.Nonnull;
 import javax.imageio.ImageIO;
+import javax.swing.*;
```
DethRaid: What does Swing give us?

mod/target/.history
```diff
@@ -0,0 +1,4 @@
+exit()
```
DethRaid: What's this file for?

native/.gitignore
```diff
@@ -1,5 +1,13 @@
 # Only things for native itself in here! src has its own .gitignore
 
+<<<<<<< HEAD
```
DethRaid: ...

Why are you trying to merge a git merge conflict?

DethRaid (changes requested):

DethRaid (approved):

PR #144, opened 2018-07-21
Auto Init Submodules In Gradle

ckissane:


mod/build.gradle
```diff
@@ -405,7 +418,7 @@ task copyReleaseBuilds(dependsOn: buildReleaseNatives) {
 processResources.mustRunAfter copyReleaseBuilds
 build.dependsOn copyReleaseBuilds
 
-task buildDevLibrary() {
+task buildDevLibrary(dependsOn: updateSubmodules) {
```
DethRaid, line 421: Will this try to update the submodules every time? How will gradle know that the submodules are already downloaded?

DethRaid (changes requested):

mod/build.gradle
ckissane: git knows so its fine

ckissane: (I tested)

DethRaid (approved):

PR #147, opened 2018-07-21
Do not crash on exit anymore

Janrupf:
Note that fixing the crash on exit revealed a lot of validation errors

### DIFF
--- a/native/src/utils/crash_handler.cpp
+++ b/native/src/utils/crash_handler.cpp
@@ -1,0 +1,60 @@
+//
+// Created by jannis on 19.07.18.
+//
+
+#include "crash_handler.h"
+
+#include <csignal>
+#include <execinfo.h>
+#include <iostream>
+#include <exception>
+#include <typeinfo>
+
+namespace nova {
+    void crash_handler::install() {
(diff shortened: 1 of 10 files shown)


native/src/utils/crash_handler.h
```diff
@@ -0,0 +1,21 @@
+//
+// Created by jannis on 19.07.18.
+//
+
+#ifndef NOVA_RENDERER_CRASH_HANDLER_H
+#define NOVA_RENDERER_CRASH_HANDLER_H
+
+#include <csignal>
```
DethRaid: Should move this to the .cpp file since it's not used here

DethRaid (changes requested):

DethRaid (approved):

native/src/utils/crash_handler.cpp
```diff
@@ -0,0 +1,60 @@
+//
+// Created by jannis on 19.07.18.
+//
+
+#include "crash_handler.h"
+
+#include <csignal>
+#include <execinfo.h>
```
DethRaid, line 8: I tried running this in a branch I'm working on and I get a compile error that this header can't be found

PR #150, opened 2018-07-29
Changes from master forge vulkan

Niko-sk2x:
This should be all that there was in master, and also some more because porting the changes dirrectly just didn't work.

Currently there are still some issues with some renderchunks rendering in the wrong places, I don't know what causes it. And it sometimes seems to cause a segfault. Other than that, it mostly works.


Niko-sk2x:
Now for some reason it still segfaults on shutdown. Should I try to debug it?

mod/run/shaderpacks/DefaultShaderpack/shaders/gui.frag
```diff
@@ -59,12 +59,12 @@ layout(location = 0) out vec4 color_out;
 
 void main() {
     if(textureSize(colortex, 0).x > 0) {
-        vec4 tex_sample = texture(colortex, uv);
+        vec4 tex_sample = texture(colortex, mod(uv,1.0));
```
DethRaid, line 62: The texture coords should already be in the 0 - 1 range though... right?

DethRaid (approved):

mod/run/shaderpacks/DefaultShaderpack/shaders/gui.frag
Niko-sk2x: I think someone added it so that the menu background is actually tiled, and I think the proper way to do that would be to actually make the textures tiled when exceeding uv 0-1, instead of clamping the value (which I'm pretty sure is an option)

PR #151, opened 2018-08-04
Nova debug replay

DethRaid:
Merging in the debug replay branch. This tool is super useful and I love it

I've also made some bugfixes and reorganizations on this branch and I realized I needed to get those in. Basically, Nova was creating a new staging buffer and a new command buffer every time a texture was updated, and was creating a new command buffer each frame. These things by themselves were just kinda silly and not awful, but the fact that Nova was destroying these things meant it had to use fences to wait for the command buffers to finish executing, which lead to a GPU deadlock and a timeout. This no longer happens in the debug replay tool.

I also made GUI geometry follow the same code path as chunk geometry. This let me put the geometry updates at the beginning of the frame, which removed a VkWaitForFence call - furthering my goal of removing fences

Finally, I added a preprocessor definition `DUMP_NOVA_CALLS` which enables the debug dumping. The CMake file enables this for regular Nova but disables it for the debug replay. This paves the way for not having to copy the debug dump file, but I didn't make that change.


native/src/CMakeLists.txt
```diff
@@ -224,6 +225,7 @@ source_group("nova" FILES ${NOVA_HEADERS} ${NOVA_NO_COMPILE} ${NOVA_SOURCE})
 
 add_library(nova-renderer SHARED ${NOVA_SOURCE} ${SPIRV_CROSS_SOURCE})
 set_target_properties(nova-renderer PROPERTIES POSITION_INDEPENDENT_CODE TRUE)
+target_compile_definitions(nova-renderer PUBLIC DUMP_NOVA_CALLS)
```
Niko-sk2x: might be a good idea to have it enabled only in DEBUG build

native/src/mc_interface/mc_interface_dump.h
```diff
@@ -0,0 +1,255 @@
...
+}
+
+void dump_destruct() {
+    dump_lock.lock();
+    DUMP_FUNC;
+    dump_lock.unlock();
+    dump_out.close();
+}
+
+#else
```
Niko-sk2x: I know the way you did it is less code, but with that big #ifded ... #else, every time you add something to the mc interface, you also need to make sure it's added to both branches of this ifdef. Consider only doing the #ifdef on the function body, not all the functions. More code, but it's not as easy to break it without noticing it.

native/src/render/objects/resources/texture2D.cpp
```diff
@@ -255,6 +261,9 @@ namespace nova {
 
         command_buffer.copyBufferToImage(buffer, image, vk::ImageLayout::eTransferDstOptimal, 1, &region);
 
+#pragma clang diagnostic push
+#pragma ide diagnostic ignored "OCDFAInspection"
```
Niko-sk2x: what...

Niko-sk2x (commented):
I don't really understand most of the vulkan changes that well, so I can't see anything wrong. So I will leave that for someone else to review.

native/src/render/objects/resources/texture2D.cpp
DethRaid: I don't like warnings, and this one is a false positive. I can remove it if you want

native/src/CMakeLists.txt
DethRaid: Fixed

native/src/mc_interface/mc_interface_dump.h
DethRaid: Fixed

native/src/render/objects/resources/texture2D.cpp
DethRaid: Fixed

native/src/mc_interface/mc_interface_dump.h
```diff
@@ -16,18 +16,30 @@
 #define W_FLOAT(x) do { float d[] = { (float) x }; dump_out.write((char*) d, sizeof(float)); } while(false)
 #define W_DOUBLE(x) do { double d[] = { (double) x }; dump_out.write((char*) d, sizeof(double)); } while(false)
 
+#else
+
+#define DUMP_FUNC
+#define W_UINT32(x)
+#define W_FLOAT(x)
+#define W_DOUBLE(x)
```
Niko-sk2x: And do these even need to be defined if not used?

DethRaid: Oh, are they only used in one file? I wasn't sure

DethRaid: Fixed

DethRaid:
About the Vulkan changes: I'm probably the one who knows Vulkan the best - I'm not sure anyone else could review them. Could you download and run this branch? If it runs without any new errors I'll be satisfied

Niko-sk2x:
I'm getting a lot of those:
```
[08:30:56] [Nova Renderer/ERROR]: ERROR: API: Validation Object: VK_NULL_HANDLE (Type = 12) | Cannot get query results on queryPool 0xe3 with index 3 as data has not been collected for this index.
```

On this line:
```
            vk::Result query_readback_result = context->device.getQueryPoolResults(context->timestamp_query_pool, 0, buffer.size() , buffer.size() * sizeof(uint64_t), buffer.data(), sizeof(uint64_t), vk::QueryResultFlagBits::e64);
```

in `nova_renderer::end_frame`.

PR #152, opened 2018-08-04
Debug replay merge

DethRaid:
- Adds a tool to capture all Nova API calls, then replay them in a JVM-free environment
- Tool is only enabled on debug builds of the Nova shared library
- Compiles on Windows


Janrupf:
Reopening for CI build to trigger

Niko-sk2x (commented):
Looks good. One thing I think could be changed is move the decoding of the dump file to the same file as the encoding, to keep it all in one place.

DethRaid:
That would be a good change but I don't think it should delay merging in the debug replay

PR #155, opened 2018-08-18
[DO NOT MERGE]Add macOS support

QianNangong:
It still need some hacks to make native build on macOS works.
Xcode bundled clang doesn't support C++17 std::filesystem, but cmake will always select it instead of `g++7`.


Janrupf:
Looks good. Please note that we are currently completly rewriting nova. It would still be cool if you finish this PR.

QianNangong:
It's really hard to track why it doesn't work.
Close for now.

DethRaid:
Could you define "doesn't work"?

QianNangong:
It can't show a window, and the renderer won't work as on Windows or Linux.
I closed it for now because I have less time to check whether it is a bug of MoltenVK or not.
When I have time, I will rework on it.

PR #171, opened 2019-01-29
Fix unused entries in .gitmodules

cwfitzgerald:
It broke jenkins, though it doesn't take a lot to do that....

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -28,3 +28,6 @@
 [submodule "3rdparty/miniz"]
 	path = 3rdparty/miniz
 	url = https://github.com/richgel999/miniz.git
+[submodule "3rdparty/minitrace"]
+	path = 3rdparty/minitrace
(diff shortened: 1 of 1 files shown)


.gitmodules
```diff
@@ -28,9 +28,3 @@
 [submodule "3rdparty/miniz"]
 	path = 3rdparty/miniz
 	url = https://github.com/richgel999/miniz.git
-[submodule "3rdparty/enkiTS"]
-	path = 3rdparty/enkiTS
-	url = https://github.com/dougbinks/enkiTS.git
-[submodule "3rdparty/minitrace"]
```
DethRaid: Minitrace should be in there

DethRaid (changes requested):

.gitmodules
cwfitzgerald: Sheeeit

DethRaid (approved):

PR #176, opened 2019-02-05
Fix image layouts

DethRaid:
Fix validation errors about image layouts by inserting image memory barriers in the appropriate places

I've added code to the shaderpack loading code. The new code examines the render graph and find edges between image usage - edges where one render pass writes to a texture that a previous render pass reads from, or vice versa. I also have special barriers for the swapchain images because they need special love


CMakeLists.txt
```diff
@@ -310,6 +309,7 @@ if (CMAKE_BUILD_TYPE STREQUAL "Debug" AND NOT NOVA_IN_SUBMODULE)
     set(NOVA_TEST ON)
 endif()
 
+set(NOVA_TEST ON)
```
Janrupf: Thats not good, add at least a check if nova is in a submodule.

src/render_engine/vulkan/vulkan_render_engine.cpp
```diff
@@ -1874,20 +1893,17 @@ namespace nova {
         }
 
         // Find where we are in the list of passes
-        auto itr = render_passes_by_order.rbegin();
-        while(itr != render_passes_by_order.rend()) {
-            if(*itr == pass.data.name) {
+        int32_t idx = 0;    // Signed so it can be negative so the `for` loop will exit
```
Janrupf: Doesn't compile with `-Werror` due to signed/unsigned compare

Janrupf (changes requested):

Janrupf:
Very nice, can confirm that all validation erros are gone

CMakeLists.txt
DethRaid: We need to talk about this actually. The way you were setting NOVA_TEST doesn't work at all for msvc. Is there an easy way to support msvc or do I have to pass the variable on the command line?

Janrupf: I'm passing the variable to the cmake command line, better said Clion does it for me.

cwfitzgerald:
test this please

Janrupf (approved):

PR #177, opened 2019-02-05
Fix all MSVC errors

DethRaid:
Fix all the errors that MSVC gave me with our current warning settings

I did not increase any warning levels or enable any more warnings


DethRaid:
Test this please

src/render_engine/dx12/dx12_render_engine.cpp
```diff
@@ -489,7 +489,7 @@ namespace nova {
                 }
                 texture->SetName(s2ws(texture_name).c_str());
 
-                auto new_tex_index = dynamic_textures.size();
+                const uint32_t new_tex_index = static_cast<uint32_t>(dynamic_textures.size());
```
cwfitzgerald, line 492: Why not use auto when there's a static cast on the right had side.

src/render_engine/vulkan/vulkan_render_engine.cpp
```diff
@@ -520,11 +520,11 @@ namespace nova {
     std::future<uint32_t> vulkan_render_engine::add_mesh(const mesh_data& input_mesh) {
         return scheduler->add_task(
             [&](ttl::task_scheduler*, const mesh_data* input_mesh, std::mutex* mesh_upload_queue_mutex, std::queue<mesh_staging_buffer_upload_command>* mesh_upload_queue) {
-                const uint32_t vertex_size = input_mesh->vertex_data.size() * sizeof(full_vertex);
-                const uint32_t index_size = input_mesh->indices.size() * sizeof(uint32_t);
+                const uint32_t vertex_size = static_cast<uint32_t>(input_mesh->vertex_data.size()) * sizeof(full_vertex);
```
cwfitzgerald, line 523: Why not auto?

cwfitzgerald (changes requested):

src/render_engine/dx12/dx12_render_engine.cpp
DethRaid: I don't like auto

src/render_engine/vulkan/vulkan_render_engine.cpp
DethRaid: Not a huge fan of auto if it won't save many characters

cwfitzgerald (approved):

PR #179, opened 2019-02-06
Make Codebase Clean with Werror

cwfitzgerald:
Fixes various warnings and build system issues that were causing warnings. Merging this is of pretty high importance as the jenkins config now uses ./jenkins.sh to build, so if that's not found, that's an issue.


Janrupf (approved):
Looks good

src/render_engine/vulkan/vulkan_render_engine.cpp
```diff
@@ -141,6 +141,7 @@ namespace nova {
         depth_barriers.reserve(textures.size());
 
         for(const auto& [_, texture] : textures) {
+            (void) _;
```
DethRaid, line 144: What and why is this?

cwfitzgerald: The code is using `_` to be representative of an unused value, but c++ doesn't special case `_` as an unused value. Therefore you must void cast it to tell the compiler to stfu. If you want to change this back to a normal variable name that's fine, but was out of scope of this PR.

PR #181, opened 2019-02-07
Mesh loading

DethRaid:
- Mesh loading happens before rendering
- Renderpass fences start in signaled state
    - Because mesh upload waits for all renderpass fences
- Mesh loading no longer tries to use the task scheduler because there is no task scheduler


3rdparty/3rdparty.cmake
```diff
@@ -109,4 +109,4 @@ set_property(TARGET spirv-cross-glsl PROPERTY EXCLUDE_FROM_ALL True)
 set_property(TARGET spirv-cross-hlsl PROPERTY EXCLUDE_FROM_ALL True)
 set_property(TARGET spirv-cross-msl PROPERTY EXCLUDE_FROM_ALL True)
 set_property(TARGET spirv-cross-reflect PROPERTY EXCLUDE_FROM_ALL True)
-set_property(TARGET spirv-cross-util PROPERTY EXCLUDE_FROM_ALL True)
+set_property(TARGET spirv-cross-util PROPERTY EXCLUDE_FROM_ALL True)
```
cwfitzgerald: But why tho

cwfitzgerald (changes requested):
Idk about the true meaning, but seams fine from a code standpoint... besides that one newline you removed for no good reason.

3rdparty/3rdparty.cmake
DethRaid: git

cwfitzgerald (approved):

PR #184, opened 2019-02-08
Split vk render engine

DethRaid:
I split the Vulkan render engine across a few files because tbh it _is_ a few files


src/render_engine/vulkan/vulkan_render_engine.hpp
```diff
@@ -30,10 +29,9 @@
 
 #include "../../render_objects/render_object.hpp"
 #include <mutex>
+#include "swapchain.hpp"
```
cwfitzgerald, line 32: Include order, but that's currently unenforced, so ignoring.

cwfitzgerald (approved):
Sure

PR #186, opened 2019-02-12
Clang Tidy and Clang Format

cwfitzgerald:
This is a monster PR. 75 commits. In an effort to make this readable, I have split this into commits by change type. A brief guide for reviewing this:

- Commits starting with `clang-tidy:` are a direct result of the named clang-tidy check. There is no more than one of these checks per commit.
- The commits names `THICC: Apply automatic clang-tidy-fixes` and `reformat` are very large, programmatic changes. These should most likely be skimmed, as reading the entire thing will bore you to death. If there is a problem with formatting, that is outside of the scope of this PR.
- Windows is broken. It will be fixed shortly. Windows headers are extremely include order sensitive.

Major Changes:
- Added `format` `reformat`, `lint`, `lint-by-file`, `lint-by-diagnostic`, and `lint-fix` targets.
- Boosted warnings up to `-Wall -Wextra -Wpedantic -Wconversion -Werror`
- Added fmtlib for formatting. Full integration of fmtlib comes later.
- Moved gtest into 3rd party folder to prevent warning fall through.
- Automatic build failure on lack of formatting or warnings from any compiler.


.clang-tidy
```diff
@@ -1,10 +1,10 @@
...
+Checks:          'boost*,bugprone*,cert*,clang-analizer-*,clang-analyzer-*,clang-diagnostic-*,cppcoreguidelines*,google*,hicpp-multiway-paths-covered,llvm*,misc*,modernize*,performance*,readability*,-*-avoid-magic-numbers,-*-dcl59-cpp,-*-err58-cpp,-*-header-guard,-*-magic-numbers,-*-move-const-arg,-*-non-private-member-variables-in-classes,-*-pro-bounds-constant-array-index,-*-pro-bounds-pointer-arithmetic,-*-pro-type-reinterpret-cast,-*-pro-type-union-access,-*-readability-todo,-*-runtime-int,-*-runtime-references,-*-string-constructor'
 WarningsAsErrors: ''
 HeaderFilterRegex: ''
 AnalyzeTemporaryDtors: false
 FormatStyle:     none
-User:            nova
+User:            connor
```
DethRaid, line 7: Shouldn't the user still be Nova? Alternately, what's the meaning of the User variable?

.gitmodules
```diff
@@ -1,5 +1,5 @@
 [submodule "tests/3rdparty/googletest"]
-	path = tests/3rdparty/googletest
+	path = 3rdparty/googletest
```
DethRaid, line 2: We had put googletest under the tests directory because only the tests depend on it

src/debugging/renderdoc.cpp
```diff
@@ -3,22 +3,22 @@
 #include "../platform.hpp"
 #include "../util/logger.hpp"
 
-#if NOVA_WINDOWS
+#if defined(NOVA_WINDOWS)
 #include "../util/windows.hpp"
 #include "../util/windows_utils.hpp"
 
 // Fucking hell
 #ifdef ERROR
```
DethRaid, line 11: Why didn't you change this one to `if defined()`?

src/debugging/renderdoc.cpp
```diff
@@ -3,22 +3,22 @@
...
 #undef ERROR
 #endif
-#elif NOVA_LINUX
-#include <dlfcn.h>
+#elif defined(NOVA_LINUX)
 #include "../util/linux_utils.hpp"
+#include <dlfcn.h>
```
DethRaid, line 15: I thought system headers usually went first

src/debugging/renderdoc.cpp
```diff
@@ -3,22 +3,22 @@
...
 #endif
 
 namespace nova {
-    RENDERDOC_API_1_3_0* load_renderdoc(const std::string& renderdoc_dll_path) {
-#if NOVA_WINDOWS
+    RENDERDOC_API_1_3_0 *load_renderdoc(const std::string &renderdoc_dll_path) {
```
DethRaid: Actually the * and & should be on the left, next to the type, following with the C++ core guidelines

src/loading/folder_accessor.cpp
```diff
@@ -7,19 +7,21 @@
 #include "../util/logger.hpp"
 
 namespace nova {
-    folder_accessor_base::folder_accessor_base(fs::path folder) : root_folder(std::make_shared<fs::path>(folder)), resource_existence_mutex(new std::mutex) {}
+    folder_accessor_base::folder_accessor_base(const fs::path &folder)
+        : root_folder(std::make_shared<fs::path>(folder)), resource_existence_mutex(new std::mutex) {
+    }
```
DethRaid: Shouldn't put the closing curly brace on a new line if the body is empty

src/loading/folder_accessor.hpp
```diff
@@ -6,30 +6,31 @@
...
         const std::string message;
         const std::error_code error_code;
 
     public:
-        explicit filesystem_exception(const fs::filesystem_error &error) : message(error.what()), error_code(error.code()) {}
+        explicit filesystem_exception(const fs::filesystem_error &error) : message(error.what()), error_code(error.code()) {
+        }
 
-        const char *what() const noexcept override {
+        [[nodiscard]] const char *what() const noexcept override {
```
DethRaid: What does nodiscard mean and why do we need it?

src/loading/json_utils.hpp
```diff
@@ -80,7 +80,9 @@ namespace nova {
      * \return An optional that contains the value, if it can be found, or an empty optional if the value cannot be found
      */
     template <typename ValType>
-    std::optional<ValType> get_json_value(const nlohmann::json &json_obj, const std::string &key, std::function<ValType(const nlohmann::json &)> deserializer) {
+    std::optional<ValType> get_json_value(const nlohmann::json &json_obj,
```
DethRaid: I feel like this line has an absurd amount of trailing whitespace but I can't tell from a GitHub diff on mobile

src/loading/shaderpack/json_interop.hpp
```diff
@@ -23,15 +23,15 @@ namespace nova {
...
 
     void from_json(const nlohmann::json &j, render_pass_data &pass);
 
     void from_json(const nlohmann::json &j, std::vector<render_pass_data> &passes);
 
     void from_json(const nlohmann::json &j, material_pass &pass);
 
-    void from_json(const nlohmann::json &j, material_data &pass);
-}  // namespace nova
+    void from_json(const nlohmann::json &j, material_data &mat);
```
DethRaid: Good catch!

DethRaid (changes requested):
Looks like the clang-format file still needs a little tweaking

.clang-tidy
cwfitzgerald: Yes, fixed in later commit.

It's used for one check that forces you to add your name to todos:

```
// TODO(connor): Blah
```

But that's off, so doesn't matter.

.gitmodules
cwfitzgerald: I know, but for a slew of reasons, 3rd party deps really need to be self contained within /3rdparty. That being said, I can turn off the googletest import if you aren't building tests.

src/debugging/renderdoc.cpp
cwfitzgerald: Fixed later.

src/debugging/renderdoc.cpp
cwfitzgerald: That's the way clang-format/tidy like them. There's no real solid argument one way or the other as you shouldn't need any system headers for your own headers. Shouldn't:tm:. There is probably a reason why they decided that way in the first place, but I don't know it.

src/debugging/renderdoc.cpp
cwfitzgerald: Fixed.

src/loading/folder_accessor.cpp
cwfitzgerald: Fixed.

src/loading/folder_accessor.hpp
cwfitzgerald: `nodiscard` warns if you ignore the return value. Void cast counts as a use, so if you really want to override it on a case by case basis you can, but it's a good default to have.

src/loading/json_utils.hpp
cwfitzgerald: Doesn't

.gitmodules
DethRaid: Turning it off if you're not building tests would be wonderful

src/debugging/renderdoc.cpp
DethRaid: And there's no setting? Ugh

src/loading/json_utils.hpp
DethRaid: Probably just the mobile website being weird

src/debugging/renderdoc.cpp
cwfitzgerald: Wait, no it's fully customizable, you set regexes for an include and give it a category number, things go in category order, then alpha within category. What is your preferred order?

src/loading/folder_accessor.cpp
cwfitzgerald: Your choice seems to be that it considers that function as non empty, so you either get all one-line-able functions on one line, or you get stuff like this here and there.

src/debugging/renderdoc.cpp
```diff
@@ -3,22 +3,22 @@
 #include "../platform.hpp"
 #include "../util/logger.hpp"
 
-#if NOVA_WINDOWS
-#include "../util/windows.hpp"
-#include "../util/windows_utils.hpp"
+#if defined(NOVA_WINDOWS)
+#    include "../util/windows.hpp"
```
DethRaid: I don't know how I missed this the first time I reviewed but can we not separate the # and the token? Like, I have never done this in my code and I don't know why clang-format thought it was okay

DethRaid (changes requested):

DethRaid (approved):

PR #187, opened 2019-02-21
Ubos

DethRaid:
Adds in UBOs. Yay!

`uniform_buffer_object` is a dumb UBO. You give it some data, it uploads the data, big whoop.

`auto_buffer` is a large buffer that allocates itself into smaller bits. Original use case is allocating space on the GPU for model matrices for render objects, but I expect it'll get used a lot more in the future

I also moved the whole renderer from namespace `nova` to namespace `nova::renderer`. This will help us keep the renderer contained when Nova gets developed into a full engine

I tweaked the `clang-tidy` settings to remove the warning about using `auto`. `auto` is fine sometimes but I really don't like it. I come to C++ for explicit types!


src/render_engine/vulkan/auto_allocating_buffer.cpp
```diff
@@ -0,0 +1,160 @@
...
+                    index_to_allocate_from = i;
+                }
+            }
+        }
+
+        VkDescriptorBufferInfo ret_val;
+
+        if(index_to_allocate_from == -1) {
+            // Whoops, couldn't find anything
+            auto ss = std::stringstream{};
```
cwfitzgerald: Favor fmtlib

src/render_engine/vulkan/auto_allocating_buffer.cpp
```diff
@@ -0,0 +1,160 @@
...
+        }
+
+        auto& chunk_to_allocate_from = chunks[static_cast<uint32_t>(index_to_allocate_from)];
+        if(chunk_to_allocate_from.range == size) {
+            // Easy: unallocate the chunk, return the chunks
+
+            ret_val = VkDescriptorBufferInfo{buffer, chunk_to_allocate_from.offset, chunk_to_allocate_from.range};
+            chunks.erase(chunks.begin() + index_to_allocate_from);
+            // NOLINTNEXTLINE(cppcoreguidelines-avoid-goto)
+            goto end;
```
cwfitzgerald: But why tho

cwfitzgerald (changes requested):
Some minor things, but looks good to my non-vulkan eyes

src/render_engine/vulkan/auto_allocating_buffer.cpp
DethRaid: Oh yeah I forgot about that.

Fixed

src/render_engine/vulkan/auto_allocating_buffer.cpp
DethRaid: Fixed

cwfitzgerald (approved):

PR #193, opened 2019-03-28
Direct rendering

DethRaid:
Indirect rendering is complex af and causes a driver crash

Direct rendering is simple and easy and _works_

Q.E.D.


cwfitzgerald (approved):
Meh, looks fine to me

PR #194, opened 2019-03-29
Render objects

DethRaid:
Render things!

This PR adds the notion of a `renderable`. A `renderable` is any object that can be rendered. Currently only static meshes are supported

Renderables are automatically batched according to their render pass, material, and mesh - e.g. Nova automatically instances multiple `renderable`s that use the same material pass and mesh. Future work will likely instance across materials, but that's not supported _yet_

I also removed the megamesh buffer and the indirect rendering because that caused my OS to crash. Meshes now have their own buffers, because it's simpler. I may go back to putting multiple meshes in the same buffer - but honestly I don't see much value at this point, since the buffers would have to be organized by renderpass and material pass and that's a lot of complexity


DethRaid:
[static_mesh_renderable](https://trello.com/c/0hwBKLes/74-staticmeshrenderable)

DethRaid:
[Uniform buffer for model matrix](https://trello.com/c/GNNN0Kze/70-uniform-buffer-for-model-matrix)

src/render_engine/vulkan/vulkan_utils.hpp
```diff
@@ -9,52 +9,28 @@
...
         if(result != VK_SUCCESS) {                                                                                                         \
-            throw exception(std::string(__FILE__) + ":" + std::to_string(__LINE__) + "=> " +                                               \
-                            ::nova::renderer::vulkan::vulkan_utils::vk_result_to_string(result));                                          \
+            NOVA_LOG(ERROR) << __FILE__ << ":" << __LINE__ << "=> " << #expr << "=" << vk_result_to_string(result);                        \
         }                                                                                                                                  \
-    } while(false);
+    }
 #else
-#define NOVA_THROW_IF_VK_ERROR(expr, exception) expr
+#define NOVA_CHECK_ERROR(expr) expr
```
Dinolek: NOVA_CHECK_ERROR is unused and compilation fail when CMAKE_BUILD_TYPE=Release.
Isn't it supposed to have same name when debug is disabled?

DethRaid: Yes - good catch!

DethRaid:
There's a bug in here where each material's renderables will overwrite the previous material's renderables' model matrices. I'm counting the slot in the model matrix buffer to write to per-material instead of globally

Will fix as soon as I get time

external/run-clang-tidy/run-clang-tidy.py
```diff
@@ -0,0 +1,477 @@
+#!/usr/bin/env python3
```
cwfitzgerald, line 1: And you're yelling at me for copying a file into foundational 😛

src/debugging/renderdoc.cpp
```diff
@@ -17,22 +18,21 @@
...
 #if defined(NOVA_WINDOWS)
-        HINSTANCE const renderdoc_dll = LoadLibrary(renderdoc_dll_path.c_str());
+        using HINSTANCE = HINSTANCE__* const;
+        HINSTANCE renderdoc_dll = LoadLibrary(renderdoc_dll_path.c_str());
         if(!renderdoc_dll) {
             const std::string error = get_last_windows_error();
-            NOVA_LOG(ERROR) << "Could not load RenderDoc. Error: " << error;
-
-            return nullptr;
+            return result<RENDERDOC_API_1_3_0*>(MAKE_ERROR("Could not load RenderDoc. Error: {:s}", error));
```
cwfitzgerald, line 27: You could probably drop the template argument if you added some deduction guides so that CTAD worked.

cwfitzgerald (approved):
Nothing stuck out as particularly heinous, but take a look about CTAD in result

PR #195, opened 2019-03-30
Each material doesn't overwrite model matrices

DethRaid:
In Nova, all the model matrices are stored in one large model matrix buffer

May or may not eventually switch to an instance-rate vertex buffer, but that's a separate issue

When recording drawcalls, the old code would write model matrices to the model matrix buffer by starting each material at index 0. This meant each material's model matrices would overwrite the previous materials' model matrices. This change moves from a local index to a global one, so each successive material writes to a new index

There are come considerations for when Nova becomes threaded, these are expressed in the comments this PR adds


Janrupf (approved):
Looks good and works as expected

PR #197, opened 2019-03-31
Create a proper `include/` directory

DethRaid:
I've created a `include/` directory in the root of the repo for public headers. These are included with `<>` in the rest of the code - I've changed all the files that include them

I've also removed all usages of the task scheduler, since we're not actually using it. We can add task scheduling back in when the time comes


cwfitzgerald (approved):
Sehr gut

PR #200, opened 2019-04-10
Clean include

Janrupf:
This PR should fix some issues with includes (mostly these fixes are just for keeping them sorted better).
~~This PR won't fix the mixed use of `#pragma once` and include guards, we'll have to discuss whats better later.~~

EDIT: Will change to `#pragma once` for everything


.clang-format
```diff
@@ -59,16 +59,16 @@ ForEachMacros:
   - foreach
   - Q_FOREACH
   - BOOST_FOREACH
-IncludeBlocks:   Preserve
+IncludeBlocks:   Regroup
```
cwfitzgerald, line 62: Nah, this needs to be preserve in order to separate out problematic headers.

include/nova_renderer/nova_renderer.hpp
```diff
@@ -1,12 +1,8 @@
 /*!
  * \brief Main class for Nova. This class exists as a singleton so it's always available
```
cwfitzgerald: If you're going to keep this it should be on the nova-renderer class.

cwfitzgerald (changes requested):

.clang-format
Janrupf: Not sure about that, in my opinion we should try to not depend on include order at all. At the moment it works, are you sure that should be changed back?

cwfitzgerald: I agree we should try not to and none of the headers we write will be, but some really really dumb headers exist that are not self-sufficient.

src/loading/shaderpack/json_interop.hpp
```diff
@@ -32,6 +26,4 @@ namespace nova::renderer {
     void from_json(const nlohmann::json& j, material_pass& pass);
 
     void from_json(const nlohmann::json& j, material_data& mat);
-} // namespace nova::renderer
-
-#endif // NOVA_RENDERER_JSON_INTEROP_HPP
+} // namespace nova::renderer
```
DethRaid: Missing trailing newline

src/loading/shaderpack/render_graph_builder.cpp
```diff
@@ -1,11 +1,9 @@
-/*!
- * \author ddubois
- * \date 17-Sep-18.
- */
-
 #include "render_graph_builder.hpp"
```
DethRaid, line 1: Why didn't this get sorted to the bottom of the includes?

src/loading/shaderpack/shaderpack_validator.hpp
```diff
@@ -6,12 +6,12 @@
  * \date 12-Sep-18.
```
DethRaid: Should remove the author/date header comment

src/render_engine/vulkan/vulkan_render_engine.cpp
```diff
@@ -1,25 +1,30 @@
...
 
+#include <spirv_cross/spirv_glsl.hpp>
+
+#include "nova_renderer/util/platform.hpp"
+
+#include "../../loading/shaderpack/render_graph_builder.hpp"
+#include "../../loading/shaderpack/shaderpack_loading.hpp"
+#include "../../util/logger.hpp"
+
+// TODO: Move windowing out of render engine folders to prevent things like this
```
DethRaid: Please define "things like this"

DethRaid: And say why they're bad

src/util/logger.hpp
```diff
@@ -1,19 +1,14 @@
...
 
 #include <functional>
 #include <mutex>
 #include <sstream>
 #include <string>
 #include <unordered_map>
+#include <utility>
 
+// Hope this fixes windows build for now, not sure where it comes from
+// TODO: Fix the root of this issue (@dethraid, @cwfitzgerald)
```
DethRaid: The root of the issue is that windows.h is stupid. We need to undefine ERROR

DethRaid: Please change these comments to something like "windows.h defines ERROR, and Nova uses ERROR as a lot level, so we need to undefine ERROR so we can it"

DethRaid (changes requested):

src/loading/shaderpack/render_graph_builder.cpp
Janrupf: Because its the header belonging to the source file

DethRaid (approved):

cwfitzgerald (approved):

PR #204, opened 2019-05-02
[DO NOT MERGE] Example for Language Translation System

skrekhere:
DO NOT MERGE, UNFINISHED.
just looking for more ideas on it, but it will be useful in the future

### DIFF
--- a/src/loading/shaderpack/shaderpack_validator.cpp
+++ b/src/loading/shaderpack/shaderpack_validator.cpp
@@ -2,7 +2,9 @@
 
 #include <fmt/format.h>
 
-#include <WinNls.h>
+#include <locale>
+
+#include <fstream>
 
 #include "nova_renderer/util/utils.hpp"
 
@@ -43,7 +45,13 @@ namespace nova::renderer {
 
     nlohmann::json default_texture_format = {{"pixelFormat", "RGBA8"}, {"dimensionType", "Absolute"}};
 
-	LANGID defaultLanguage = GetUserDefaultLangID();
+	
+    std::ifstream ifs("./../../utils/lang_master.json");
+    nlohmann::json j = nlohmann::json::parse(ifs);
+
+	std::locale l("");
+    std::string local = l.name();
+    
 
     void ensure_field_exists(nlohmann::json& j,
                              const std::string& field_name,
--- a/src/loading/shaderpack/shaderpack_validator.cpp
+++ b/src/loading/shaderpack/shaderpack_validator.cpp
@@ -2,6 +2,8 @@
 
 #include <fmt/format.h>
 
+#include <WinNls.h>
+
 #include "nova_renderer/util/utils.hpp"
 
 #include "../json_utils.hpp"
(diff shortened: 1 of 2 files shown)


cwfitzgerald:
Can one of the admins verify this patch?

cwfitzgerald:
Test this please.

camatthew (dismissed):
Good idea

src/loading/shaderpack/shaderpack_validator.cpp
```diff
@@ -2,6 +2,8 @@
 
 #include <fmt/format.h>
 
+#include <WinNls.h>
```
DethRaid: What is this header?

skrekhere: It gives me the ability to use GetUserDefaultLangID()

DethRaid: That's not helpful

Where is it from? What does it provide? What other solutions have you considered and why did you go with this one?

skrekhere: I have not tried any other solutions. What it does is gives me the current user’s default language so that I can pull that from the language json. I plan to use a different solution in the future as this one is very far from cross-platform, but for now this is a temporary solution for testing, given I work in windows

skrekhere: I believe the header itself lets me pull quite a few helpful options from a windows user’s settings

DethRaid: Oh, so the proposal is more about the json file and how data is stored there? That would be a great thing to mention in the PR description

I'd recommend looking at c++ localization libraries. I don't know what's available and how well it'd fit Nova, but we won't know until we look.

foxfirecodes: Also, how would this handle on other operating systems? Wouldn't you need to special case this based on current OS so that there's a way to get the user lang for Linux as well?

foxfirecodes: Actually, it looks like this might help: https://stackoverflow.com/questions/3702665/how-can-i-get-the-system-language-in-c-c

I think `std::locale("");` should work; apparently using a nameless locale gives you the system default :)

skrekhere: I wasn’t looking for a cross platform way of doing things just yet. Right now I’m using that header as a base testing method, and I’ll do research on cross platform alternatives when I get a chance

foxfirecodes: Ah, fair enough. Well, for reference in the future:

**Linux:**
This returns the system locale, based on the `LANG` environment variable.

`setlocale(LC_ALL, NULL);`

**macOS:**
You need a specific header for macOS.

https://stackoverflow.com/questions/39255590/how-to-get-the-locale-of-the-current-user-in-osx-using-c

DethRaid:
Have you had any luck with developing this further?

skrekhere:
yeah, i have, it’s just a matter of actually implementing it everywhere in the codebase now

skrekhere:
when i get to a further point i’ll make sure to push

cwfitzgerald:
Closing because we're moving to rust with nova-rs.

PR #207, opened 2019-06-04
Vulkan Backend Improvements

ghost:
I added improvements for Vulkan API and some engine code.
- [x] Added macro option `ENABLE_VULKAN` and `ENABLE_D3D12` into code
- [x] Made some code clean and validation fixes (in VS Code, Windows 10)
- [x] Improved Vulkan API structures for better access (but I'm not sure, need save VMA, or need completely to remove/replace)


cwfitzgerald:
Can one of the admins verify this patch?

include/nova_renderer/command_list.hpp
```diff
@@ -77,7 +78,7 @@ namespace nova::renderer::rhi {
          * \pre destination_offset plus num_bytes is less than the size of source_buffer
          */
         virtual void copy_buffer(
-            Buffer* destination_buffer, uint64_t destination_offset, Buffer* source_buffer, uint64_t source_offset, uint64_t num_bytes) = 0;
+            struct Buffer* destination_buffer, uint64_t destination_offset, struct Buffer* source_buffer, uint64_t source_offset, uint64_t num_bytes) = 0;
```
DethRaid, line 82: I'd rather forward declare things than have an inline tentative declaration like this

include/nova_renderer/nova_renderer.hpp
```diff
@@ -325,7 +327,7 @@ namespace nova::renderer {
         void record_material_pass(MaterialPass& pass, rhi::CommandList* cmds);
 
         void record_rendering_static_mesh_batch(MeshBatch<StaticMeshRenderCommand>& batch, rhi::CommandList* cmds);
-#endif
+#pragma endregion
```
DethRaid: I don't see a change from #if to #pragma region

include/vulkan/vk_icd.h
```diff
@@ -0,0 +1,175 @@
+//
```
DethRaid: Why are you including this file into Nova?

src/render_engine/vulkan-hpp/README.md
```diff
@@ -0,0 +1,4 @@
+# Vulkan HPP Backend
```
DethRaid: But why? Why does Nova need two Vulkan backends, one with the C header and one with the C++ header?

src/render_engine/vulkan/swapchain.hpp
```diff
@@ -10,8 +10,9 @@
 #include <glm/glm.hpp>
 #include <vector>
 #include <vulkan/vulkan.h>
-#include "nova_renderer/util/utils.hpp"
+#include <nova_renderer/util/utils.hpp>
```
DethRaid: We made the decision to always use double quotes for any file that's part of Nova, and to use <> for external files

src/render_engine/vulkan/vk_structs.hpp
```diff
@@ -7,38 +7,73 @@
...
+#include <vma/vk_mem_alloc.h>
+#include <vulkan/vulkan.h>
+#include <nova_renderer/rhi_types.hpp>
 
+#ifdef ENABLE_VULKAN
 namespace nova::renderer::rhi {
+
     struct VulkanDeviceMemory : DeviceMemory {
-        VkDeviceMemory memory;
+        VmaAllocationInfo info{};
```
DethRaid: No

The whole point of DeviceMemory is to enable Nova to manage its own gpu memory. DirectX has no equivalent for VMA, so we could have to do custom memory stuff for that backend. I realized that what Nova needed from GPU memory wasn't super complicated, so I decided to manually do memory management for Vulkan as well

Much of the Vulkan render engine does still use VMA, yes, but only because this branch is very incomplete. VMA will be completely removed before my bra ch is merged into master

include/nova_renderer/device_memory_resource.hpp
```diff
@@ -3,16 +3,16 @@
...
 
 namespace nova::renderer {
-    namespace rhi {
-        struct DeviceMemory;
-    }
 
     struct DeviceMemoryAllocation {
-        rhi::DeviceMemory* memory = nullptr;
+        //rhi::DeviceMemory* memory = nullptr;
+        rhi::DeviceMemory memory{}; // I suppose should be it owner!
```
DethRaid: This won't work... Like at all

First off, Nova will have a lot of DeviceMemoryAllocations for a single DeviceMemory. The DeviceMemoryAllocation doesn't own the DeviceMemory

Even if it did own the DeviceMemory it was allocated from, DeviceMemory is a struct that API-specific structs inherit from. Nova uses pointers for all the RHI structs to allow each RHI backend to store its own data for an image, descriptor set, etc. We shouldn't use value types anywhere for RHI structs

DethRaid (changes requested):
High level thoughts:
* I like the operator* on all the Vulkan structs. Using that to access the Vulkan objects is very nice
* The preprocessor guards around each RHI backend are very appreciated
* Please don't put headers for external code in Nova. We use VCPKG for dependency management, any dependencies should be handled by that 
* Why does Nova need a vulkan.hpp backend? I'd rather facilitate custom RHI backends by having NovaRenderer accept an RHI backend as a constructor parameter than have community-made RHI backends in the Nova repository
* Nova has a .clang-format file, please use it 

Overall I really appreciate your initiative and there's definitely some good ideas in here, but please talk with us so you can work on things Nova needs

include/nova_renderer/rhi_types.hpp
```diff
@@ -37,7 +38,7 @@ namespace nova::renderer::rhi {
     };
 
     struct Buffer : Resource {
-        uint64_t size = 0;
+        uint64_t offset = 0u, size = 0u; // added offset, because Vulkan API may use different address in allocated memory
```
DethRaid: The offset for Vulkan memory allocations (or any other allocations) is stored in a `DeviceMemoryAllocation` struct

If it wasn't, Vulkan-specific data for buffers should be stored in `VulkanBuffer`, which already has some Vulkan-specific data such as the `VkBuffer` itself

src/render_engine/vulkan/vk_structs.hpp
```diff
@@ -7,30 +7,39 @@
...
+#include "nova_renderer/rhi_types.hpp"
 
 #ifdef ENABLE_VULKAN
 namespace nova::renderer::rhi {
 
     struct VulkanDeviceMemory : DeviceMemory {
-        VmaAllocationInfo info{};
-        VmaAllocation allocation{};
-        VmaMemoryUsage usage = VMA_MEMORY_USAGE_GPU_ONLY;
+        //VmaAllocationInfo info{};
```
DethRaid: Delete things, don't comment them out and leave them in there

DethRaid (changes requested):

PR #208, opened 2019-10-17
GitHub actions

Janrupf:
Here we go, CI is back but this time with the brand new Github Actions feature.
This workflow relies on the `setup-*` scripts and thus is currently suitable for the master branch, but as soon as `command_buffer` gets merged this needs to be updated.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,300 @@
...
+    # Even tho we run the real build in an arch linux docker
+    # containerwe need to specify a HostOS
+    runs-on: ubuntu-latest
+
+    # Run the real build on an arch linux docker container
+    # Ubuntu has out of date tools and sometimes system updates
+    # require manual interaction so its unusable for CI
+    container: archlinux/base
+
+    steps:
```
DethRaid: Looks like a lot of this is similar to the GCC steps. Can we extract each step into a GitHub action?

DethRaid (changes requested):

.github/workflows/ci.yml
Janrupf: This is confirmed very similar, however, Github Actions themself are not simple bash scripts but entire javascript steps which do not run in the container specified by the build job, so this is not practical for our use case.

cwfitzgerald: You should be able to pull common functionality into a meta-step or meta-job or something so you don't duplicate it.

Janrupf: As far as I know there is no such functionality. You can't define step and reuse them. Maybe also because of the fact this is still beta. For now I think this is the best we can get. Feel free to point out if there is a possibility. It could be possible to let the vcpkg build only run once on one system and share that via an artifact, but that would add more steps than it would remove (since system update and such would then be duplicated 3 times).

cwfitzgerald: Damn, azure has it but github actions doesn't. I honestly would sooner use a build matrix, as build matrices seem a bit more sane in github actions than in travis.

DethRaid (approved):

.github/workflows/ci.yml
```diff
@@ -0,0 +1,300 @@
...
+      # 3. Checkout the repository, this is done via a github action
+      - name: Checkout repository
+        uses: actions/checkout@master
+      # 4. Prepare nova
+      #    Currently there are 2 different build methods in use with nova,
+      #    so check which one the branch we are on is using and adjust as required
+      - name: Prepare
+        shell: bash
+        run: |
+          set -e
```
cwfitzgerald: This is redundant, gha already sets it

.github/workflows/ci.yml
```diff
@@ -0,0 +1,300 @@
...
+              # Now vcpkg should be set up
+              VULKAN_SDK=/usr bash setup-linux.sh
+          fi
+      # 5. Configure with CMake
+      - name: Configure with CMake
+        run: |
+          mkdir build
+          cd build
+          VULKAN_SDK=/usr cmake ..   \
+            -G"Unix Makefiles"       \
```
cwfitzgerald: Use ninja

.github/workflows/ci.yml
```diff
@@ -0,0 +1,300 @@
+name: CI
+on: [push]
```
cwfitzgerald: Prs too

.github/workflows/ci.yml
```diff
@@ -0,0 +1,300 @@
...
+          cp build/nova-test-end-to-end artifacts
+          cp build/libnova-renderer.a artifacts
+      # 9. Upload binary artifacts
+      - name: Upload binary artifacts
+        uses: actions/upload-artifact@master
+        with:
+          name: binary-linux-gcc
+          path: artifacts
+
+  build-linux-clang:
```
cwfitzgerald: This duplication can be solved with a build matrix. See https://github.com/BVE-Reborn/bve-reborn/blob/master/.github/workflows/build.yml for how I did it across OSs, you can use the same technique for across compilers.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,300 @@
...
+    # Running on Windows Server 2019 or whatever the latest version is,
+    # lets hope this goes well...
+    runs-on: windows-latest
+
+    steps:
+      # 1. Download the VulkanSDK
+      - name: Download VulkanSDK
+        shell: powershell
+        run: |
+          $web = New-Object System.Net.WebClient
```
cwfitzgerald: Wtf just call curl

cwfitzgerald: This whole section should just be a call to curl.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,300 @@
...
+      # 3. Checkout the repository, this is done via a github action
+      - name: Checkout repository
+        uses: actions/checkout@master
+      # 4. Prepare nova
+      #    Currently there are 2 different build methods in use with nova,
+      #    so check which one the branch we are on is using and adjust as required
+      - name: Prepare
+        shell: bash
+        run: |
+          set -e
```
cwfitzgerald: Redundant

.github/workflows/ci.yml
```diff
@@ -0,0 +1,300 @@
...
+        with:
+          name: docs
+          path: artifacts/docs
+      # Upload the headers artifact
+      - name: Upload headers
+        uses: actions/upload-artifact@master
+        with:
+          name: headers
+          path: artifacts/headers
+
```
cwfitzgerald: Empty line with whitespace

cwfitzgerald (changes requested):
In general good, a couple issues.

cwfitzgerald (approved):

PR #209, opened 2019-10-20
Render Hardware Interface

DethRaid:
This PR rewrites the renderer into two components. There's the Render Hardware Interface (RHI), which presents a low-level abstraction over D3D12, Vulkan, and OpenGL. There's also the API-agnostic rendering code, which is written entirely in terms of the RHI. This will allow us to easily add in new RHI backends without having to make a whole new renderer


cwfitzgerald:
Note to self: do not override, do not merge before testing `./nova-end-to-end` on both windows and linux.

Janrupf (approved):
Vulkan backend works on Linux and unit tests pass, looks good to me

PR #211, opened 2019-10-29
Pull Documenation from nova-rs and de-rustify it

cwfitzgerald:
May not be perfect, but much better.


DethRaid (approved):

cwfitzgerald:
Overriding due to github actions being a mess currently.

PR #213, opened 2019-11-12
Fix typo in README

cwfitzgerald:


DethRaid (approved):
Only thing is that I wasn't aware Ninja was the preferred build system :p

PR #214, opened 2019-11-14
Add test shaderpack

DethRaid:
Adds back the test shaderpack, and removes the line in the .gitignore file that ignores the test shaderpack


cwfitzgerald (changes requested):
Hello, it's Sargent Nitpick, basically every file is missing a newline at the end of it.

cwfitzgerald (approved):

PR #215, opened 2019-11-14
More buffer support

DethRaid:
This PR gives the RHI the ability to bind a buffer to a descriptor. This will enable game-custom UBOs and shaderpack-custom UBOs

Additionally, I moved the two builtin buffers - NovaModelMatrixUBO and NovaPerFrameUBO - to a `std::unordered_map`. This allows me to grab the builtin buffers directly through their name, instead of through something like a massive `switch` statement

Finally, I refactored some of the code for finding something in a `std::unordered_map` to use `if` statements with initializers


include/nova_renderer/constants.hpp
```diff
@@ -0,0 +1,10 @@
+//! \brief A collection of constants for Nova to use
+
+#pragma once
+
+#include <string>
+
+namespace nova::renderer {
+    const std::string ModelMatrixBufferName = "NovaModelMatrixBuffer";
+    const std::string PerFrameDataName = "NovaPerFrameUBO";
```
Janrupf: Variable naming seems a bit wrong here. I think this should either be `SCREAMING_SNAKE_CASE` or `snake_case`.

src/render_engine/dx12/dx12_render_engine.cpp
```diff
@@ -654,7 +654,7 @@ namespace nova::renderer::rhi {
         return semaphore;
     }
 
-    std::vector<Semaphore*> D3D12RenderEngine::create_semaphores(uint32_t num_semaphores) {
+    std::vector<Semaphore*> D3D12RenderEngine::create_semaphores(const uint32_t num_semaphores) {
```
Janrupf, line 657: Not sure if `const` makes sense here in any way. Since this is not a reference and moreover a primitive type, I don't see a reason using const here unless you want to stop yourself from accidently assigning this a new value.

Janrupf (changes requested):
Looks pretty good, except those 2 things. Tested end-to-end with Vulkan backend and works.

src/render_engine/dx12/dx12_render_engine.cpp
DethRaid: Partly I want to keep myself from assigning, partly I like to use `const` whenever possible

Janrupf: Ok, as discussed on discord, lets use const as much as possible, so this can stay there for sure.

include/nova_renderer/constants.hpp
DethRaid: Fixed

Janrupf (approved):

cwfitzgerald (commented):
Nothing terribly wrong here.

cwfitzgerald (approved):

PR #218, opened 2019-11-21
Switch from incrementally reading with `ifstream` to reading the whole thing with `fread`

DethRaid:
`ifstream` is gross and bad. `fread` is not. QED

### DIFF
--- a/src/loading/regular_folder_accessor.cpp
+++ b/src/loading/regular_folder_accessor.cpp
@@ -1,7 +1,5 @@
 #include "regular_folder_accessor.hpp"
 
-#include <fstream>
-
 #include "../util/logger.hpp"
 
 namespace nova::renderer {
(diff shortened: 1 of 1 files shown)


src/loading/regular_folder_accessor.cpp
```diff
@@ -20,28 +18,28 @@ namespace nova::renderer {
...
-        while(getline(resource_stream, buf)) {
-            // uint8_t val;
-            // resource_stream >> val;
-            // buf.push_back(val);
-            file_string += buf.c_str();
-            file_string += "\n";
-        }
+        std::fseek(resource_file, 0, SEEK_SET);
+
+        std::fread(file_string.data(), sizeof(char), file_size, resource_file);
```
cwfitzgerald, line 40: This isn't a problem, but I find this hilarious as `sizeof` is literally defined as "how many chars this thing takes up". :stuck_out_tongue:

cwfitzgerald (approved):
LGTM

src/loading/regular_folder_accessor.cpp
DethRaid: I could hardcode this but idk, seems lame

PR #221, opened 2019-12-02
Procedural mesh class

### LINKED ISSUE #223: Add Define for Debug Checks Separate from NDEBUG
Per #221:

Two main reasons for this, optimization and debug checks are orthogonal and you may want an optimized build with all checks on and `#ifndef NDEBUG` is terrible for readability as it's a double negative. `#if NOVA_DEBUG` is much more clear.

DethRaid:
I've added a procedural mesh class. This allows one to upload new mesh data every frame, facilitating highly dynamic geometry 

Currently ProceduralMesh uploads its data every frame. This is probably fine, but future work will profile and optimize so mesh data isn't uploaded unless needed

Procedural Meshes are owned by NovaRenderer. I created MapAccessor as a way to access an object in a map without knowing the map itself. This allows code external to NovaRenderer to set the mesh data of a procedural mesh

I've also renamed a few things to better match how the current code works - for instance, "renderable" is becoming "render command", and "render engine" is starting to become "render device" - although render device will be a separate PR

Finally, the frame index that a Swapchain returns is now a `uint8_t`. This better reflects that there will only ever be a small number of in-flight frames at a given time

Closes #223

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -344,9 +344,8 @@ set_target_properties(nova-renderer PROPERTIES POSITION_INDEPENDENT_CODE TRUE)
 target_compile_definitions(nova-renderer PUBLIC DUMP_NOVA_CALLS)
 target_compile_definitions(nova-renderer PUBLIC MTR_ENABLED)
 target_compile_definitions(nova-renderer PUBLIC FMT_STRING_ALIAS=1)
-if(CMAKE_BUILD_TYPE MATCHES DEBUG)
-    target_compile_definitions(nova-renderer PUBLIC NOVA_DEBUG)
-endif()
+target_compile_definitions(nova-renderer PUBLIC $<$<CONFIG:Debug>:NOVA_DEBUG>)
+
 target_include_directories(nova-renderer PUBLIC
         $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
         $ENV{VULKAN_SDK}/Include
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -135,6 +135,7 @@ set(NOVA_SOURCE
 		
         src/render_objects/procedural_mesh.cpp
         src/render_objects/uniform_structs.hpp
+        src/render_objects/renderables.cpp
 
         src/util/logger.cpp
         src/util/logger.hpp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,7 +36,7 @@ option(NOVA_ENABLE_VULKAN_RHI "Compile the Vulkan RHI backend" ON)
 option(NOVA_ENABLE_D3D12_RHI "Compile the D3D12 RHI backend" ON)
 option(NOVA_ENABLE_OPENGL_RHI "Compile the OpenGL 3 backend" ON)
 
-option(NOVA_FORCE_DEBUGGING "Force compiling all the debugging and validation code" ON)
+option(NOVA_FORCE_DEBUGGING "Force compiling all the debugging and validation code" OFF)
 
 if(NOVA_ENABLE_D3D12_API AND NOT WIN32)
     message(WARNING "Nova can only compile its D3D12 RHI backend on Windows 10. You're now on Windows 10, so I'm disabling the D3D12 RHI backend")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,6 +97,7 @@ set(NOVA_SOURCE
         include/nova_renderer/util/platform.hpp
         include/nova_renderer/util/result.hpp
         include/nova_renderer/util/utils.hpp
+        include/nova_renderer/util/container_accessor.hpp
 
         include/nova_renderer/command_list.hpp
         include/nova_renderer/device_memory_resource.hpp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -115,6 +115,7 @@ set(NOVA_SOURCE
         include/nova_renderer/bytes.hpp
         include/nova_renderer/polyalloc.hpp
         include/nova_renderer/constants.hpp
+        include/nova_renderer/frontend/procedural_mesh.hpp
 
         src/nova_renderer.cpp
         src/render_engine/render_engine.cpp
@@ -130,7 +131,8 @@ set(NOVA_SOURCE
         src/loading/json_utils.hpp
 
         src/settings/nova_settings.cpp
-
+		
+        src/render_objects/procedural_mesh.cpp
         src/render_objects/uniform_structs.hpp
 
         src/util/logger.cpp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -344,6 +344,9 @@ set_target_properties(nova-renderer PROPERTIES POSITION_INDEPENDENT_CODE TRUE)
 target_compile_definitions(nova-renderer PUBLIC DUMP_NOVA_CALLS)
 target_compile_definitions(nova-renderer PUBLIC MTR_ENABLED)
 target_compile_definitions(nova-renderer PUBLIC FMT_STRING_ALIAS=1)
+if(CMAKE_BUILD_TYPE MATCHES DEBUG)
+    target_compile_definitions(nova-renderer PUBLIC NOVA_DEBUG)
+endif()
 target_include_directories(nova-renderer PUBLIC
         $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
         $ENV{VULKAN_SDK}/Include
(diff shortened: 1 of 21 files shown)


CMakeLists.txt
```diff
@@ -130,8 +132,10 @@ set(NOVA_SOURCE
         src/loading/json_utils.hpp
 
         src/settings/nova_settings.cpp
-
+		
+        src/render_objects/procedural_mesh.cpp
```
cwfitzgerald, line 138: Who do you think you are, adding _source files_?

include/nova_renderer/constants.hpp
```diff
@@ -11,4 +11,6 @@ namespace nova::renderer {
     const uint32_t AMD_PCI_VENDOR_ID = 0x1022;
     const uint32_t INTEL_PCI_VENDOR_ID = 8086;
     const uint32_t NVIDIA_PCI_VENDOR_ID = 0x10DE;
+
+    const uint8_t NUM_IN_FLIGHT_FRAMES = 3;
```
cwfitzgerald: I like aeroplanes

include/nova_renderer/frontend/procedural_mesh.hpp
```diff
@@ -0,0 +1,84 @@
+#pragma once
+
+#include <array>
+#include <cstdint>
+
+#include "nova_renderer/constants.hpp"
+
+namespace nova::renderer {
+    namespace rhi {
```
cwfitzgerald, line 9: Can this be extracted into a forward declaration header as to not have to repeat and possibly get conflicts? Maybe with all of the forward decls from the rhi?

include/nova_renderer/frontend/procedural_mesh.hpp
```diff
@@ -0,0 +1,84 @@
...
+        std::array<rhi::Buffer*, NUM_IN_FLIGHT_FRAMES> vertex_buffers;
+        std::array<rhi::Buffer*, NUM_IN_FLIGHT_FRAMES> index_buffers;
+
+        rhi::Buffer* cached_vertex_buffer;
+        rhi::Buffer* cached_index_buffer;
+
+        uint64_t num_vertex_bytes_to_upload = 0;
+        uint64_t num_index_bytes_to_upload = 0;
+
+#ifndef NDEBUG
```
cwfitzgerald: We should add a custom define for Debug/Release

include/nova_renderer/nova_renderer.hpp
```diff
@@ -40,13 +42,34 @@ namespace nova::renderer {
...
+         * For example, a non-animated mesh just needs a mat4 for its model matrix
+         *
+         * This buffer gets re-written to every frame, since the number of renderables in this mesh batch might have changed. If there's
+         * more renderables than the buffer can hold, it gets reallocated from the RHI
+         */
+        rhi::Buffer* per_renderable_data = nullptr;
+
+        std::vector<RenderCommandType> commands;
+
+        ProceduralMeshBatch(std::unordered_map<MeshId, ProceduralMesh>& meshes, MeshId key) : mesh(meshes, key) {}
```
cwfitzgerald, line 64: Does MapAccessor store a reference? If so this is valid code, if not, you should be taking meshes by value and moving.

cwfitzgerald: Answered my own question, ignore.

src/nova_renderer.cpp
```diff
@@ -792,6 +811,38 @@ namespace nova::renderer {
...
+                    cur_model_matrix_index * sizeof(glm::mat4),
+                    model_matrix_buffer);
+                cur_model_matrix_index++;
+            }
+        }
+
+        if (start_index != cur_model_matrix_index) {
+        const auto& [vertex_buffer, index_buffer] = batch.mesh->get_buffers_for_frame(cur_frame_idx);
+            // TODO: There's probably a better way to do this
+            const std::vector<rhi::Buffer*> vertex_buffers = { vertex_buffer,
```
cwfitzgerald: `const std::vector<rhi::Buffer*> vertex_buffers(7, vertex_buffer);`

I always have to look up the decl for `std::vector::vector`...

src/nova_renderer.cpp
```diff
@@ -810,24 +861,55 @@ namespace nova::renderer {
...
-                    command.model_matrix = glm::translate(command.model_matrix, renderable.initial_position);
-                    command.model_matrix = glm::rotate(command.model_matrix, renderable.initial_rotation.x, {1, 0, 0});
-                    command.model_matrix = glm::rotate(command.model_matrix, renderable.initial_rotation.y, {0, 1, 0});
-                    command.model_matrix = glm::rotate(command.model_matrix, renderable.initial_rotation.z, {0, 0, 1});
-                    command.model_matrix = glm::scale(command.model_matrix, renderable.initial_scale); // Uniform scaling only
-
-                    batch.renderables.emplace_back(command);
+        StaticMeshRenderCommand command = make_render_command(renderable, id);
+
+        if(const auto itr = meshes.find(renderable.mesh); itr != meshes.end()) {
```
cwfitzgerald, line 860: Praise be if initalizers.

src/render_objects/procedural_mesh.cpp
```diff
@@ -0,0 +1,206 @@
...
+          vertex_buffer_size(vertex_buffer_size),
+          index_buffer_size(index_buffer_size)
+#endif
+    {
+        const auto aligned_vertex_buffer_size = bvestl::polyalloc::align(vertex_buffer_size, 256);
+        const auto aligned_index_buffer_size = bvestl::polyalloc::align(index_buffer_size, 256);
+        const auto host_memory_size = aligned_vertex_buffer_size + aligned_index_buffer_size;
+        const auto device_memory_size = host_memory_size * 3;
+
+        // TODO: Don't allocate a separate device memory for each procedural mesh
```
cwfitzgerald: s/a separate/separate/

src/render_objects/procedural_mesh.cpp
```diff
@@ -0,0 +1,206 @@
...
+                return true;
+            })
+            .on_error([](const ntl::NovaError& error) {
+                NOVA_LOG(ERROR) << "Could not allocate host memory for procedural mesh. Error: " << error.to_string();
+                // TODO: Propagate the error
+            });
+    }
+
+    void ProceduralMesh::set_vertex_data(const void* data, const uint64_t size) {
+#ifndef NDEBUG
```
cwfitzgerald: This double negative is really hard to read. See above.

cwfitzgerald (changes requested):
Some commentary, a couple nits, and some general sleepy sass.

include/nova_renderer/constants.hpp
cwfitzgerald: Actually this should be constexpr, like the rest of the consts.

include/nova_renderer/frontend/procedural_mesh.hpp
DethRaid: Why? `NDEBUG` is standard since C++98

include/nova_renderer/frontend/procedural_mesh.hpp
DethRaid: I'm not sure you put this comment on the right line

include/nova_renderer/frontend/procedural_mesh.hpp
cwfitzgerald: Two main reasons, optimization and debug checks are orthogonal and you may want an optimized build with all checks on. Aditionally `#ifndef NDEBUG` is terrible for readability as it's a double negative. `#if NOVA_DEBUG` is much more clear.

include/nova_renderer/frontend/procedural_mesh.hpp
cwfitzgerald: That whole namespace should be lifted into a forward decl, which is why I put it there.

DethRaid: Isn't it already a forward declaration, though?

Or do you want a separate header for RHI forward declarations? Cause I'd rather just forward declare what I need

include/nova_renderer/constants.hpp
DethRaid: The rest of the constants in this file are `const`

Why should it be `constexpr`? Should everything in this file be `constexpr`? Why not `const`?

include/nova_renderer/frontend/procedural_mesh.hpp
cwfitzgerald: Yeah I want it in a separate file so the rhi file can include it and verify the forward decls are correct.

include/nova_renderer/constants.hpp
cwfitzgerald: That was a poorly worded instruction to make them all constexpr. The strings really should be `std::string_view` which is constexpr compatible. `(static) const int` is special at top level to act like constexpr. This is a remnant of the C days and the inconsistency is trash.

cwfitzgerald:
So this fucking broke because std::string_view isn't the same as std::string and I want to scream as this would be easily possible in rust, but c++ be playin.

include/nova_renderer/frontend/procedural_mesh.hpp
DethRaid: I'll make a forward declaration header in another PR. Don't let me forget

include/nova_renderer/frontend/procedural_mesh.hpp
cwfitzgerald: Gonna upgrade this to an issue.

cwfitzgerald (approved):
Upgrading forward decls and the define to issues.

CMakeLists.txt
```diff
@@ -36,6 +36,8 @@ option(NOVA_ENABLE_VULKAN_RHI "Compile the Vulkan RHI backend" ON)
 option(NOVA_ENABLE_D3D12_RHI "Compile the D3D12 RHI backend" ON)
 option(NOVA_ENABLE_OPENGL_RHI "Compile the OpenGL 3 backend" ON)
 
+option(NOVA_FORCE_DEBUGGING "Force compiling all the debugging and validation code" ON)
```
cwfitzgerald: This needs to default off.

cwfitzgerald (changes requested):
So close

cwfitzgerald:
Closes #223

cwfitzgerald (approved):
LGTM

DethRaid:
Merging because the failure looks like the CI push script being weird and we don't want to be blocked by push, only pull-request

PR #227, opened 2019-12-05
Better input callbacks

DethRaid:
Add callbacks for the mouse position and buttons, and add more button information to the keyboard input callback

### DIFF
--- a/include/nova_renderer/window.hpp
+++ b/include/nova_renderer/window.hpp
@@ -38,15 +38,20 @@ namespace nova::renderer {
          * This callback will key called for every key press event that this window receives
          *
          * \param key_callback Callback for when a key is received. Intentionally a std::function so I can easily add
-         * it to a vector. First parameter to this function is the key code, second is whether the key was pressed this frame, third is the
-         * modifier key, if any
+         * it to a vector. First parameter to this function is the key code, second is whether the key was pressed
+         * this frame, third is if control is down, fourth is if shift is down
          */
-        void register_key_callback(std::function<void(uint32_t, bool, uint32_t)>&& key_callback);
+        void register_key_callback(std::function<void(uint32_t, bool, bool, bool)>&& key_callback);
 
         /*!
-         * \brief Sends the provided key to all registered key callbacks
+         * \brief Registers a new mouse position callback
+         *
+         * Mouse position callback gets called when the mouse position changed
+         *
+         * \param mouse_callback Callback for when mouse input is received. The first parameter is the mouse's X
+         * position, the second if the Y position
          */
-        void process_key(int key);
+        void register_mouse_callback(std::function<void(double, double)>&& mouse_callback);
 
         void poll_input() const;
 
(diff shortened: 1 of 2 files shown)


include/nova_renderer/window.hpp
```diff
@@ -72,6 +88,22 @@ namespace nova::renderer {
...
+
+        std::vector<std::function<void(uint32_t, bool)>> mouse_button_callbacks;
+
+        static void glfw_key_callback(GLFWwindow* window, int key, int scancode, int action, int mods);
+
+        static void glfw_mouse_callback(GLFWwindow* window, double x_position, double y_position);
+
+        static void glfw_mouse_button_callback(GLFWwindow* window, int button, int action, int mods);
+
+        void broadcast_key_event(int key, bool is_press, bool is_control_down, bool is_shift_down);
```
cwfitzgerald, line 103: Should be `is_pressed`

src/windowing/window.cpp
```diff
@@ -18,15 +18,34 @@ typedef int Bool;   // Because X11 is stupid
...
+
+    void NovaWindow::glfw_mouse_callback(GLFWwindow* window, const double x_position, const double y_position) {
+        void* user_data = glfwGetWindowUserPointer(window);
+        auto* my_window = static_cast<NovaWindow*>(user_data);
+
+        my_window->broadcast_mouse_position(x_position, y_position);
+    }
+
+    void NovaWindow::glfw_mouse_button_callback(GLFWwindow* window, const int button, const int action, int /* mods */) {
+		void* user_data = glfwGetWindowUserPointer(window);
```
cwfitzgerald: Identation

include/nova_renderer/window.hpp
```diff
@@ -38,14 +38,30 @@ namespace nova::renderer {
          * This callback will key called for every key press event that this window receives
          *
          * \param key_callback Callback for when a key is received. Intentionally a std::function so I can easily add
-         * it to a vector
+         * it to a vector. First parameter to this function is the key code, second is whether the key was pressed
+         * this frame, third is if control is down, fourth is if shift is down
          */
-        void register_key_callback(std::function<void(uint32_t)>&& key_callback);
+        void register_key_callback(std::function<void(uint32_t, bool, bool, bool)>&& key_callback);
```
cwfitzgerald: Why are you taking these by rvalue reference? This probably should just be by value, as you want both rvalues and lvalues to be able to call it, and you're going to copy it anyway.

cwfitzgerald (changes requested):
Went through it with the finest comb I have, nothing really major, some minor issues, but should be quickly fixable.

include/nova_renderer/window.hpp
DethRaid: idk it seemed kinda fun

cwfitzgerald: One note, when you do write it with a copy (or rather by value), the insertion into the vector needs to be a move.

cwfitzgerald (approved):
LGTM

PR #229, opened 2019-12-16
[BestFriend] UI support work

DethRaid:
I've made a `Renderpass` class

It has a virtual method

I have no regrets

A renderpass class with a virtual method means it's easy for Nova to make buitin passes that accomplish a specific task. The PR contains one such example: a builtin pass for rendering UI. Nova expects that the host application will further subclass the UI pass to accomplish UI rendering

This PR is marked "do not merge" because I need to test out that workflow. I'll use this branch in the mesh viewer and, once I have successfully rendered a UI, I'll remove the do not merge label from here. This means that this PR will probably stick around for a while. I'm okay with that


DethRaid:
Closed until I have the branch where I want it

PR #231, opened 2019-12-27
Add libmono to Build System

cwfitzgerald:
This PR simply adds CI and CMake machinery for the use of mono in nova. There are probably more issues that need to be solved, but those will come after we start using it.


external/external.cmake
```diff
@@ -72,6 +72,65 @@ include(minitrace)
...
 
+# Mono
+if (WIN32)
+	set(MONO_LIB_NAME libmono-static-sgen)
+else() 
+	set(MONO_LIB_NAME monosgen-2.0)
+endif()
+if (MONO_ROOT)
+	message("${MONO_ROOT}/lib")
+	message("${MONO_ROOT}/include/mono-2.0")
```
Janrupf: Those debug message should probably be removed

Janrupf (changes requested):

external/external.cmake
cwfitzgerald: Yup, that escaped me.

Janrupf (approved):

cwfitzgerald:
Do not merge until deth reviews

.github/workflows/ci.yml
```diff
@@ -102,17 +102,29 @@ jobs:
           -o VulkanSDK-Installer.exe
       # 2. Install the VulkanSDK
       - name: Install VulkanSDK
+        shell: powershell
```
DethRaid, line 105: Yasssss

cwfitzgerald: 🙄 😛

README.md
```diff
@@ -34,6 +34,11 @@ We are still in early development of the multi-backend rewrite and things may ch
 
 [Contributing](docs/contributing.md).
 
+### Dependencies
+
+Two dependencies need to be installed system wide. The [LunarG Vulkan SDK](https://vulkan.lunarg.com/sdk/home) and [mono](https://www.mono-project.com/). Instructions for installation are on their websites.
```
DethRaid: I don't really know if mono should be a system dependency. I'm not sure how hard it would be to build ourselves so this is fine for now

external/external.cmake
```diff
@@ -72,6 +72,65 @@ include(minitrace)
...
+	find_library(
```
DethRaid: You are responsible for maintaining this part of the CMake file

README.md
Janrupf: We probably don't want to build it on our own, this would take a very long time...

cwfitzgerald: It's fine on linux, but on windows it's a compound msvc (for c++) + cygwin (for c#) build which sounds extraordinarily annoying.

DethRaid (changes requested):
The mono CMake stuff should be in its own file. once that is done we can merge this

external/external.cmake
cwfitzgerald: lmao alright

README.md
DethRaid: Alright. In that case let's just say the specific version of mono that we're using

DethRaid (approved):

README.md
cwfitzgerald: Done.

cwfitzgerald:
Closing due to security concerns.

PR #233, opened 2020-01-11
Resource Manager Refactor

DethRaid:
Oh lawdy

- `DeviceResource` manages device resource. You create a resource with a create info struct. There's a few kinds of resource you can create. You can then get the resources later on and use them however you want
- `Rendergraph` - same thing, but for resources. Additionally, instead of compiling the render graph into submission order at load time, we now compile it every frame. This allows me to be more dynamic about what passes exist each frame, which made Best Friend's UI code cleaner. I expect that a lot of optimizations will happen while compiling the render graph, but almost none of that code exists yet
- `PipelineManager` - same thing, but for pipelines

Renderpasses no longer have the pipelines they use backed into them. Instead, the get the pipelines they want at runtime. This allows C++ renderpasses to use renderpack pipelines, and renderpack renderpasses to use C++ pipelines. This in turn made Best Friend's UI code cleaner

Pipeline create info doesn't specify the vertex fields. Instead, Nova uses shader reflection to ask the SPIR-V module what vertex fields it needs. There's an implicit assumption that the host application and the mesh code will have some good way to specify a vertex format, and some validation that it matches the pipeline, but that code doesn't exist and won't exist for a while

I did some additional work on the allocation subsystem. I'm still not super happy with it, but what I have now works reasonably alright

There's a ton of little bugfixes in here. I fixed problems as I found them, and I found all kinds of things


cwfitzgerald (approved):
To be completely honest, this PR is way to big in scope to actually review. I have tried my best to look at it, but I probably missed a lot of shit as my eyes glazed over by file 30.

cwfitzgerald:
If there's anything in particular you want me to review, let me know, but for now I'm going to go with a general "it's probably good enough and if tests pass that's fine" stance

PR #235, opened 2020-01-21
Reorganize and rename the RHI code to better reflect what I want it to do

DethRaid:
- Rename `RenderEngine` to `RenderDevice`
- Move the implementation files from folder` render_engine` to `rhi`


cwfitzgerald (approved):
This is the perfect PR to add some really subtle change that creates a backdoor. All the changes were the same thing so I was basically skimming the color of the changes to make sure it looked right.

LGTM

PR #236, opened 2020-01-22
DeviceResources can create uniform buffers

DethRaid:
DeviceResources can create and provide access to uniform buffers


cwfitzgerald (approved):
LGTM

PR #237, opened 2020-01-23
Add Rex Core library

DethRaid:
Adds the Rex core library as a dependency of Nova

The commit history is weird but the diffs are fine


cwfitzgerald (approved):
LGTM

PR #238, opened 2020-01-24
Convert filesystem files to use Rex::Core

DethRaid:
This PR converts all the files in `include/nova_renderer/filesystem` and `src/filesystem` to use the Rex core library instead of the C++ standard library

The branch as a whole absolutely will not compile, but the files in those folders absolutely should - and they absolutely should not be using anything from the C++ standard library


src/filesystem/zip_folder_accessor.cpp
```diff
@@ -3,196 +3,206 @@
...
-        }
-
+    void print_file_tree(const FileTreeNode& folder, const uint32_t depth) {
         std::stringstream ss;
         for(uint32_t i = 0; i < depth; i++) {
             ss << "    ";
         }
 
-        ss << folder->name;
+        ss << folder.name.data();
```
cwfitzgerald, line 175: Note for the future: this should be an overload you write so you don't have to do the whole data thing.

cwfitzgerald (approved):
LGTM

src/filesystem/zip_folder_accessor.cpp
DethRaid: I'll want some string builder class tbh. Not really sure what form it'll take, though

cwfitzgerald: Makes sense. I would say to go the rust/rex way with a format string.

DethRaid: Only thing is I don't know how many thing I'll need to add to the format string ahead of time, so idk how long to make it? At some point I'll have to dynamically build a string and it'll be gross no matter what

PR #245, opened 2020-01-28
Converts the procedural mesh class to use Rex types

DethRaid:
Depends on PR https://github.com/NovaMods/nova-renderer/pull/244

PR #246, opened 2020-01-28
Remove Nova's host memory allocators

DethRaid:
This PR converts the files in `include/nova_renderer/memory` and `src/memory` to use `rx::memory::allocator` instead of `nova::mem::AllocatorHandle`. It does not touch the files outside of those folders, that's coming in the next PR

Depends on PR https://github.com/NovaMods/nova-renderer/pull/245


cwfitzgerald (approved):
LGTM

cwfitzgerald:
Overriding due to rex conversion

PR #247, opened 2020-01-28
FINISH FIRST PASS OF REX CONVERSION

DethRaid:
This PR finishes the first pass at converting Nova from the C++ standard library to Rex::Core

This PR is expected to pass CI

I repeat

THIS PR IS EXPECTED TO PASS CI

The tests pass on my local Windows machine. There may or may not be Linux-specific errors that I didn't think to check for, but CI will tell us

Next pass of the conversion will remove `fmtlib` and `nlohmann::json` to further reduce our dependency on the C++ standard library

After that I'll make a general cleanup pass because this poor code base has suffered a lot of abuse at my hands

Depends on https://github.com/NovaMods/nova-renderer/pull/246


cwfitzgerald:
Rebasing on top of master.

cwfitzgerald (approved):
LGTM

PR #248, opened 2020-01-29
Remove Fmtlib

DethRaid:
This pull request removes Nova's dependency on `fmtlib`, replacing it with calls to `rx::string::format`


cwfitzgerald (changes requested):
The non-fmt-submodule changes need to be reverted, as this will move them back to the old versions.

Otherwise LGTM

cwfitzgerald (approved):

cwfitzgerald (approved):

PR #249, opened 2020-02-01
Remove nlohmann::json

DethRaid:
Removes our dependency on nlohmann::json, replacing it with Rex's JSON parser


cwfitzgerald (approved):
LGTM

PR #251, opened 2020-02-03
Rex bugfixes

DethRaid:
Fix a bunch of bugs from me poorly using Rex


include/nova_renderer/util/utils.hpp
```diff
@@ -34,6 +34,16 @@ namespace nova::renderer {
...
     bool ends_with(const std::string& string, const std::string& ending);
 
+    template<typename Arg>
+    uint32_t va_count(const Arg& arg) {
+        return 1;
+    }
+
+    template<typename Arg, typename... Args>
+    uint32_t va_count(const Arg arg, const Args&... args) {
+        return 1 + va_count(args);
```
cwfitzgerald: should be:

```cpp
template<typename... Args>
uint32_t va_count(const Args&... args) {
    return sizeof...(args);
}
```

src/filesystem/filesystem_helpers.cpp
```diff
@@ -0,0 +1,10 @@
+#include "nova_renderer/filesystem/filesystem_helpers.hpp"
+
+#include <stdarg.h>
+#include <string.h>
+
+namespace nova::filesystem {
+    void copy_path_parts_to_buffer(const rx_size offset, char* buffer, const rx::string& path_part) {
```
cwfitzgerald: Does this need to check if the buffer is large enough?

include/nova_renderer/util/utils.hpp
cwfitzgerald: Tbh not sure this function needs to exist

src/rhi/vulkan/vulkan_render_device.cpp
```diff
@@ -1248,8 +1248,8 @@ namespace nova::renderer::rhi {
                 return compute_family_index;
         }
 
-        logger(rx::log::level::k_error, "Unknown queue type &u", static_cast<uint32_t>(type));
-        return 999999; // Will probably cause a crash, which is actually what we want rn
+        logger(rx::log::level::k_error, "Unknown queue type %u", static_cast<uint32_t>(type));
+        return 999999; // Will cause a crash, which is actually what we want rn
```
cwfitzgerald: Then either assert false, or abort don't return garbage

cwfitzgerald (changes requested):

include/nova_renderer/util/utils.hpp
DethRaid: It actually doesn't. I thought I could be cool and make a variadic template to build a path out of an arbitrary number of strings, but then msvc hit the template recursion depth and I never tried ever again

src/rhi/vulkan/vulkan_render_device.cpp
DethRaid: ugh fine I'll be reasonable

src/filesystem/filesystem_helpers.cpp
DethRaid: Deleted the function

cwfitzgerald (approved):

PR #253, opened 2020-02-24
Render things

DethRaid:
This is massive and I'm sorry

Some major changes I remember off the top of my head:

- Renderpacks only supply the rendergraph for the 3D scene. The host application supplies the render graph for the UI. Nova itself has some tasks that run before the 3D scene or UI, and some tasks the run after the 3D scene and UI
- Fixed up a lot of things in the Vulkan backend, such as semaphore creation
- Nova uses shader reflection to get the vertex data layout for each pipeline. There's an assumption that the host application will provide mesh data in the correct format, but I don't have any validation for this assumption
- Mesh data is now a void* to the vertex data and a void* to the index data. Nova no longer defines anything about the vertex format, so don't fuck up your vertices
- Better C++ API for creating renderpasses and pipelines, which Nova uses to make the render pass to layer the UI on top of the 3D scene
- RHI pipeline create info struct that's separate from the renderpack pipeline data struct. The RHI struct has all the data a pipeline needs, and some data that Nova doesn't use yet, while the renderpack struct is more user-friendly and a little higher level
- Got minitrace working and added some minitrace events to a few places in the code. A future PR will add in more profiling
- Lots of updates to rex::core as graphite developed new features and fixed bugs
- Switch from custom GPU memory allocators to VulkanMemoryAllocator


include/nova_renderer/rhi/pipeline_create_info.hpp
```diff
@@ -0,0 +1,430 @@
...
+
+        /*!
+         * \brief The state of the rasterizer when this pipeline state is active
+         */
+        RasterizerState rasterizer_state{};
+
+        rx::optional<MultisamplingState> multisampling_state{};
+
+        /*!
+         * What depth operations to perform
```
cwfitzgerald: Missing `\brief`

cwfitzgerald (changes requested):
One nit, but otherwise fine

include/nova_renderer/rhi/pipeline_create_info.hpp
DethRaid: Fixed

cwfitzgerald (approved):

cwfitzgerald:
Overriding due to incorrect vulkan version requirements

PR #258, opened 2020-03-11
Add Rex Console

DethRaid:
This PR moves the Rex core library out of the repo and into a submodule, and it makes a few changes for updated Rex APIs

This allows us to A) have smaller PRs and B) use Rex console variables eventually when I get around to it


cwfitzgerald (approved):
yas

PR #262, opened 2020-03-25
Refactor renderpasses into SceneRenderpass and GlobalRenderpass

DethRaid:
I've done a refactor

Previously, there was just one renderpass type. It was kinda cool I guess - but not cool enough

Now there's multiple types

`SceneRenderpass` is a renderpass over scene objects. It uses the bindless material system to access material parameters and textures, and it uses the standard pipeline system to render a bunch of meshes

`GlobalRenderpass` is a renderpass that operates "globally" - that is, it doesn't handle any objects in the scene. Intended use case is post-processing effects, or merging together the UI and scene output render targets. They use more bespoke code, typically using a `RhiResourceBinder` to bind resources to a pipeline

Of course, you can still create a renderpass that doesn't follow either of these paradigms. The UI renderpass is just that

What is a `RhIResourceBinder`? It's an abstraction for binding resources to pipelines. The Vulkan implementation performs shader reflection on a pipeline to extract the descriptor sets the pipelines uses, then creates `VkDescriptorSetLayout`s and a `VkPipelineLayout` to bind descriptor sets to that pipeline, and creates `VkDescriptorSet`s to actually represent the resources to the GPU. You can change bound resources at any time, and the descriptor sets won't be updated until the next time you bind a `RhiResourceBinder` to a command list

I've also split the concept of a pipeline in two - surface pipelines and global pipelines. Surface pipelines draw objects in the scene using bindless materials, global pipelines are executed by a `GlobalRenderpass` over the whole scene. Currently all pipelines loaded from a renderpack are surface pipelines - a future PR will give renderpack pipelines a way to say if they're surface or global

That's the long and the short of it


cwfitzgerald (approved):
Nothing stood out to me as terribly awful, LGTM

